PR #18, opened 2026-04-30
Draft yaml-schema

hadley:
Based on https://github.com/yaml-schema/yaml-schema/tree/main which while not documented (hence all cucumber files that claude read to produce some docs), it does have a cli (which I haven't used).

Created mostly for the purposes of discussion. Need to verify that this is something that the vscode YAML LSP uses.

@gaborcsardi the state of YAML schemas is not great 😬


hadley:
Would be better to use https://github.com/quarto-dev/q2/tree/main/crates/quarto-yaml-validation

gaborcsardi:
We chose a different direction in the end: #22.

PR #22, opened 2026-05-25
First version of schema, and first version of cli that validates a schema

gaborcsardi:


examples/dabstep.yaml
```diff
@@ -1,5 +1,7 @@
 # source: https://github.com/hadley/dabstep
 
+version: "0.1"
```
hadley: Should we make this `version: 0.1.0`?

schema.yaml
```diff
@@ -0,0 +1,172 @@
...
+    version:
+      enum: ["0.1"]
+      description: >
+        Version of the data-dict.yaml spec this document conforms to. Must be
+        quoted in YAML so it's parsed as a string, not a number.
+
+    # `name` and `description` are not in spec.md but appear in real-world
+    # files (see examples/foodbank.yaml). Allowed as optional top-level
+    # metadata; may be promoted into the spec in a later version.
+    name: string
```
hadley:
```suggestion
```

That seems like a mistake

schema.yaml
```diff
@@ -0,0 +1,172 @@
...
+                    R: string
+                    Python: string
+                    pin: string
+
+              columns:
+                arrayOf:
+                  anyOf:
+
+                    # Branch 1: type: enum  --  `values` is required.
+                    - object:
```
hadley, line 62: I think it would probably better to just have a single branch here since so much is shared between the variants. And then check that values/range/examples is correct in additional code.

crates/data-dict/tests/fixtures/invalid/bad-cardinality.yaml
```diff
@@ -0,0 +1,9 @@
+# expected: fails - `cardinality` must be one of one-to-one / one-to-many /
+# many-to-one (no many-to-many in the spec).
+
+version: "0.1"
```
hadley: Does rust have a test snapshots? It would be nice to see the validation errors for each of these (without having to run at the console, which I did)

hadley (approved):
Seems like a great start! Just a couple of small changes and then I think we can merge and iterate.

examples/dabstep.yaml
gaborcsardi: yeah

crates/data-dict/tests/fixtures/invalid/bad-cardinality.yaml
gaborcsardi: yeah, we can have snapshots

schema.yaml
gaborcsardi: You mean that it was a mistake in foodbank.yaml, which has a ‘name’ field?

schema.yaml
gaborcsardi: Yeah, I considered that. OTOH this schema does not need to be human readable. (Unlike the actual data dictionary.) So my thinking was to validate as much as possible with the YAML validator, to get uniform error messages and to avoid having to write ad-hoc code for manual schema checks.

crates/data-dict/tests/fixtures/invalid/bad-cardinality.yaml
gaborcsardi: OK, I switched to snapshots. The tricky part is that the error messages (diagrams) use some unicode characters, which might be platform dependent. We’ll see.

spec.md
```diff
@@ -1,10 +1,10 @@
 # data-dict.yaml
 
-This document describes version **0.1** of the `data-dict.yaml` specification.
+This document describes version **0.1.0** of the `data-dict.yaml` specification.
 
 A data dictionary has one required top-level key, `version`, plus three optional keys that hold the actual content:
 
-* `version` (required): the version of the `data-dict.yaml` spec this document conforms to. Currently `"0.1"`. Quote the value so it's parsed as a string, not a number.
+* `version` (required): the version of the `data-dict.yaml` spec this document conforms to. Currently `"0.1.0"`. Quote the value so it's parsed as a string, not a number.
```
hadley: Does it still need to be quoted with yaml 1.2?

gaborcsardi: No, we don’t need to. OTOH we don’t explicitly require YAML 1.2 currently, AFAICT?

schema.yaml
hadley: Hmmmm I would like it to be human readable because I want to be able to read it 😄

gaborcsardi: Hmmm. Maybe we could improve quarto-yaml-validation to be able to express this better? 

But anyway, for now, we could remove this part, but then we'll add some ad-hoc code to check all these constraints and generate error messages for them, instead just relying on the schema and our schema validator. I would actually argue that this will make the schema less readable because the constraints won't be included in it, and they'll be only expressed in the rust code of the validator.

The after-schema validation is in the other PR, so maybe this is best addressed there or after both PRs are merged. I'll do a third PR on top of #23 and then we'll see how it looks.

spec.md
gaborcsardi: done

schema.yaml
gaborcsardi: OK, I removed it.

schema.yaml
gaborcsardi: OK, this is in #25 now, this commit: https://github.com/hadley/data-dict.yaml/pull/25/changes/dbeedce89087423bef5206ad0975761eb04bd25d

gaborcsardi:
@hadley I am not sure if we can do snapshot tests, because they are platform dependent, seemingly:
https://github.com/hadley/data-dict.yaml/actions/runs/26780701900/job/79128264567

hadley:
Maybe just run snapshots on linux for now?

gaborcsardi:
> Maybe just run snapshots on linux for now?

Seems like an issue that we can fix later, it is another Unicode width measurement bug.

PR #23, opened 2026-05-25
Add cross-table semantic linting to validate-schema

gaborcsardi:
This is on top of #22.

### DIFF
--- a/crates/data-dict/src/join_expr.rs
+++ b/crates/data-dict/src/join_expr.rs
@@ -303,4 +303,124 @@ mod tests {
         let rhs = &j.conjuncts[0].rhs;
         assert_eq!(&s[rhs.start..rhs.end], "food_nutrient.fdc_id");
     }
+
+    #[test]
+    fn all_operators_parse() {
+        let cases = [
+            ("a.x = b.y", JoinOp::Eq),
+            ("a.x == b.y", JoinOp::Eq),
+            ("a.x >= b.y", JoinOp::Ge),
+            ("a.x <= b.y", JoinOp::Le),
+            ("a.x > b.y", JoinOp::Gt),
+            ("a.x < b.y", JoinOp::Lt),
+        ];
+        for (s, op) in cases {
+            let j = parse(s);
+            assert_eq!(j.conjuncts[0].op, op, "operator mismatch for {s:?}");
+        }
+    }
+
+    #[test]
+    fn two_char_operators_are_not_split() {
+        // `>=` must win over `>`, otherwise the trailing `=` would be left for
+        // the rhs qcol parse and the operator would be wrong.
+        assert_eq!(parse("a.x >= b.y").conjuncts[0].op, JoinOp::Ge);
+        assert_eq!(parse("a.x <= b.y").conjuncts[0].op, JoinOp::Le);
+    }
+
+    #[test]
+    fn whitespace_around_operator_is_optional() {
+        let j = parse("a.x=b.y");
+        assert_eq!(j.conjuncts.len(), 1);
+        assert_eq!(j.conjuncts[0].lhs.column, "x");
+        assert_eq!(j.conjuncts[0].op, JoinOp::Eq);
+        assert_eq!(j.conjuncts[0].rhs.column, "y");
+    }
+
+    #[test]
+    fn three_conjuncts() {
+        let j = parse("a.x = b.y AND c.z = d.w AND e.p = f.q");
+        assert_eq!(j.conjuncts.len(), 3);
+        assert_eq!(j.tables(), vec!["a", "b", "c", "d", "e", "f"]);
+    }
+
+    #[test]
+    fn tables_dedup_in_first_appearance_order() {
+        // `b` and `a` reappear in the second conjunct but must not be repeated.
+        let j = parse("a.x = b.y AND b.z = a.w");
+        assert_eq!(j.tables(), vec!["a", "b"]);
+    }
+
+    #[test]
+    fn qcols_yields_all_refs_in_source_order() {
+        let j = parse("a.x = b.y AND c.z = d.w");
+        let cols: Vec<(&str, &str)> =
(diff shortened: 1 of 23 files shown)


crates/data-dict/src/join_expr.rs
```diff
@@ -0,0 +1,306 @@
+//! Recursive-descent parser for the `join` expression mini-language used in
```
hadley, line 1: This is exactly what I was thinking 😄 We might also use something similar if we add richer constraints (e.g. https://github.com/hadley/data-dict.yaml/issues/10)

crates/data-dict/tests/fixtures/lint/clean-two-tables.yaml
```diff
@@ -0,0 +1,37 @@
+# expected: lints clean. Minimal two-table dict that satisfies every rule.
+
+version: "0.1"
```
hadley: Why are these in `lints/`?

crates/data-dict/tests/fixtures/lint/dd003-missing-column.yaml
```diff
@@ -0,0 +1,35 @@
+# expected: DD003
+# The relationship references `food_category.identifier`, but the column on
```
hadley, line 2: This generates three errors:

```
Error: [DD003] data-dict.yaml lint
    ╭─[ crates/data-dict/tests/fixtures/lint/dd003-missing-column.yaml:35:35 ]
    │
 35 │     join: food.food_category_id = food_category.identifier
    │                                   ────────────┬───────────  
    │                                               ╰───────────── column `identifier` is not defined in table `food_category`
────╯

Error: [DD001] data-dict.yaml lint
    ╭─[ crates/data-dict/tests/fixtures/lint/dd003-missing-column.yaml:17:15 ]
    │
 17 │       - name: food_category_id
    │               ────────┬───────  
    │                       ╰───────── column `food.food_category_id` is marked `foreign_key` but no `relationships` entry points it at a `primary_key` column
────╯

Error: [DD006] data-dict.yaml lint
    ╭─[ crates/data-dict/tests/fixtures/lint/dd003-missing-column.yaml:34:18 ]
    │
 34 │     cardinality: many-to-one
    │                  ─────┬─────  
    │                       ╰─────── cardinality is `many-to-one` but the right-side join column on `food_category` is not marked `primary_key` or `unique`
────╯
```

Is the third one redundant? (because that column doesn't exist)

crates/data-dict/tests/fixtures/lint/dd004-bad-join.yaml
```diff
@@ -0,0 +1,35 @@
+# expected: DD004
+# `food.food_category_id ~ food_category.id`: `~` is not a valid operator
+# in the join grammar.
+
+version: "0.1"
+
+tables:
+  food:
+    description: Foods.
```
hadley, line 9: I wonder if we should have some way to opt out of description, type etc, so these tests can focus on what's actually important.

crates/data-dict/tests/fixtures/lint/dd005-conflicts-not-on-both-sides.yaml
```diff
@@ -0,0 +1,39 @@
+# expected: DD005
+# `conflicts: [amount]` claims `amount` appears in both `food` and
+# `food_nutrient`, but `food` has no `amount` column.
```
hadley, line 3: Worth adding a test for the opposite? i.e. there is a conflict but it's not reported?

crates/data-dict/src/join_expr.rs
```diff
@@ -0,0 +1,306 @@
...
+//! `relationships[*].join`.
+//!
+//! Grammar:
+//!
+//! ```text
+//! join     := conjunct ("AND" conjunct)*
+//! conjunct := qcol op qcol
+//! qcol     := IDENT "." IDENT
+//! op       := "=" | "==" | ">=" | "<=" | ">" | "<"
+//! IDENT    := [A-Za-z_][A-Za-z0-9_]*
```
hadley, line 11: Probably worth testing this grammar a bit, independently of validation?

hadley (approved):
This is looking great!

crates/data-dict/tests/fixtures/lint/dd004-bad-join.yaml
gaborcsardi: Should we make the description optional in general?

crates/data-dict/tests/fixtures/lint/clean-two-tables.yaml
gaborcsardi: Because these are for the “manual” validation steps, not for the schema validation.

crates/data-dict/tests/fixtures/lint/dd003-missing-column.yaml
gaborcsardi: Indeed. Done now.

crates/data-dict/tests/fixtures/lint/dd004-bad-join.yaml
hadley: Yeah, probably. We don't want to force people to write stuff if there isn't anything to say.

gaborcsardi: done.

crates/data-dict/tests/fixtures/lint/clean-two-tables.yaml
hadley, line 16: I was thinking we could drop these too even?

gaborcsardi: column descriptions? or make all of them optional? Don’t we want to require at least a name and type?

gaborcsardi: I'll make description optional in #25 or after that.

PR #25, opened 2026-06-01
Simplify schema, remove 5 object types

gaborcsardi:
Express constraints in the validator/linter instead of the schema.

This is on top of #23.


crates/data-dict/src/lint.rs
```diff
@@ -344,6 +348,61 @@ fn side_has_unique_implied(
     })
 }
 
+// --- DD007 --------------------------------------------------------------
+
+fn check_enum_has_values(dict: &DataDict, out: &mut Vec<Diagnostic>) {
```
hadley: I think of this a bit differently: every type has some way of representing the data it contains, either a range, exhaustive set of values, or some examples. So every column should have exactly one of the three and you can tell which one based on the column type.

gaborcsardi: OK, take another look. Should the examples (and others?) be optional? It is a bit weird that `number(id)` has examples [1, 2, 3, 4, 5], especially for a foreign key.

hadley: Yeah, that's perfect. I think it is still useful to see examples there because its useful to know if its [1, 2, 3, 4, 5]` or `[10000, 1235452, 234234, 345345, 69613132]` etc.

hadley: Probably worth also looking at the spec to see if the wording could be tightened up there too.

hadley (approved):

crates/data-dict/src/lint.rs
gaborcsardi: OK, I'll do that in the next PR.

PR #30, opened 2026-06-11
Elaborate type representation in the spec

gaborcsardi:
Cf. https://github.com/hadley/data-dict.yaml/pull/25#discussion_r3374237741


hadley (approved):
Perfect, thanks!

PR #36, opened 2026-06-15
Validate examples before committing

### LINKED ISSUE #24: Validate in github action
Once #22 is in, need to update daily examples update to only update if validation passes.

hadley:
Fixes #24

### DIFF
--- a/.github/workflows/update-examples.yaml
+++ b/.github/workflows/update-examples.yaml
@@ -17,8 +17,28 @@ jobs:
         with:
           use-public-rspm: true
 
+      - uses: dtolnay/rust-toolchain@stable
+
+      - name: Build validator
+        run: cargo build -p data-dict-cli --release
+
       - name: Download examples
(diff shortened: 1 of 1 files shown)


.github/workflows/update-examples.yaml
```diff
@@ -17,9 +17,29 @@ jobs:
...
+            if ! ./target/release/data-dict validate-schema "$f"; then
+              status=1
+            fi
+            echo "::endgroup::"
+          done
+          if [ "$status" -ne 0 ]; then
+            echo "::error::One or more examples failed validation; not committing."
+            exit 1
+          fi
+
```
gaborcsardi, line 42: Do we want to deal with the case when (some of) the existing examples fail, but the updated example(s) are ok?

hadley: I think so? I'd rather find out about it quickly so I can fix it.

Issue #10, opened 2026-04-22
More constraints


hadley:
* Pattern matching (with regex)
* Conditionally missing/present (e.g. if q3 is true, then q4-8 must be present)
* Freshness (e.g. error if most recent value is >2 days old)
* String length?
* Numeric precision? (but not sure you can reliably check that)

Could we support custom expressions (using SQL syntax)? I think that would imply the creation of a table level constraints field too?

hadley:
Maybe there are potentially three places you could add constraints? Per column, per table, and per relationship? (We'd need to think through what the relationship between constraints and cardinality/join is. They seem pretty related.)

hadley:
And should we just express all these constraints as SQL-like expressions?

* `string LIKE pattern`, `string SIMILAR TO regular expression`
* `q3 = TRUE AND q6 IS NULL AND q7 IS NULL AND q8 IS NULL AND q9 IS NULL AND q10 IS NULL`. Also support list comprehensions a la duckdb? `q5 AND list_all([q6, q7, q8, q9, q10], x -> x IS NULL)`
*  `MAX(date) < now() - interval '2 weeks'`
*  `LENGTH(string) > 5`
* `PRECISION(number) <= 2`

Also need to think about if the conditions should state what is allowed or what is not allowed?

hadley:
Concrete proposal in #42

hadley:
For list columns, probably want some predicates so you can express constraints on the number of elements in each row (e.g. exactly two, at least one, no more than 5).

gaborcsardi:
Closed by #42.

PR #42, opened 2026-06-16
Propose new syntax for constraints

hadley:
#10


spec.md
```diff
@@ -125,13 +126,88 @@ The `number` type can be qualified with a measure in parentheses that classifies
...
 
+An assertion is a map with an `assert` key holding a boolean expression that must be true for every row, plus an optional `description`:
+
+```yaml
+columns:
+  - name: postcode
+    type: string
+    constraints:
+      - required
+      - assert: LENGTH(postcode) <= 10
```
lwjohnst86: Nice idea. This actually solves a problem we were trying to resolve with our [`check-datapackage`](https://check-datapackage.seedcase-project.org). If this is built into the spec, it would solve a lot of our problems trying to work around the current tooling. 👍 One worry we had was how to prevent some type of injection against if the assertion gets evaluated. But I guess that can be sanitized before evaluating.

spec.md
```diff
@@ -125,13 +126,88 @@ The `number` type can be qualified with a measure in parentheses that classifies
...
+        description: A contract can't end before it starts.
+      - assert: NOT(q3) OR (q4 IS NOT NULL AND q5 IS NOT NULL)
+        description: If q3 is true, q4 and q5 must be answered.
+```
+
+Table constraints can only carry assertions; the structural barewords (`primary_key`, `unique`, …) live on columns.
+
+### Assertions
+
+An `assert` expression is a single-table, row-level boolean expression written in a small SQL-like sublanguage. It is evaluated against every row, and the constraint holds when the expression is true for all of them. Bare names refer to columns of the table.
```
lwjohnst86: I like that it is SQL-like. SQL is basically the language of data and is so widely used in e.g. data engineering. 👍

lwjohnst86 (commented):
Really like that assertions are built into the metadata, as that would make updating and checking the data so much easier 👍

gaborcsardi:
I think it would make sense to properly formalize the grammar.

gaborcsardi:
Let's merge this and we can still refine it.

PR #48, opened 2026-06-16
Polishing validate

### LINKED ISSUE #37: `validate-schema` improvements
* Default to `data-dict.yaml` if no path.
* Can the default usage make it more clear that validate-schema requires a path? A few agents missed this.
* If version missing, suggest current version — now at https://github.com/quarto-dev/q2/pull/293
* Default usage should include link to spec — now in #46

hadley:
* Accept a directory, defaulting to current
* Try and get claude to write fewer comments and more tests

Fixes #37


.claude/claude.md
```diff
@@ -10,6 +10,11 @@ The repo contains:
 - `crates/`: Rust workspace (see crate architecture below)
 - `schema.yaml`: JSON Schema for structural validation of data dictionary files
 
+## Code principles
+
+* Reserve comment from explaining why, not what or how.
```
gaborcsardi: for?

crates/data-dict-cli/src/main.rs
```diff
@@ -72,6 +81,24 @@ fn main() -> ExitCode {
...
+                Ok(dict) => dict,
+                Err(err) => {
+                    if json {
+                        println!(
+                            "{}",
+                            serde_json::json!({
+                                "status": "error",
+                                "kind": "schema",
+                                "message": err,
+                            })
```
gaborcsardi: Should this be a "regular" error instead of a validation error? E.g. if the cli cannot read the schema file or the data file, or the parquet file is invalid, (I hope) we get a regular error, not a validation error. 

Probably makes sense to distinguish between "could not run validation" and "ran validation, and it failed".

hadley: Yeah, that makes sense.

PR #59, opened 2026-06-18
Check required columns don't have missing values

hadley:
First step to validating more values, so worth thinking a bit to lay out the right framework.

### DIFF
--- a/crates/data-dict-parquet/Cargo.toml
+++ b/crates/data-dict-parquet/Cargo.toml
@@ -7,8 +7,6 @@ license.workspace = true
 repository.workspace = true
 
 [dependencies]
-# `snap` adds Snappy decompression, needed by `null_report` to read data pages
-# (the most common parquet codec). Schema reading alone needs no codec.
 parquet = { workspace = true, features = [
   "snap",
   "flate2",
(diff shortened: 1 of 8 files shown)


crates/data-dict-parquet/Cargo.toml
```diff
@@ -7,4 +7,6 @@ license.workspace = true
 repository.workspace = true
 
 [dependencies]
-parquet.workspace = true
+# `snap` adds Snappy decompression, needed by `null_report` to read data pages
+# (the most common parquet codec). Schema reading alone needs no codec.
+parquet = { workspace = true, features = ["snap"] }
```
gaborcsardi: We might as well add the others as well because we'll need to support them eventually.

gaborcsardi:
I like the infrastructure in general. 

Specifically for required columns and parquet files, can't we just require that the column is also REQUIRED in the parquet file? Or do we want to allow OPTIONAL parquet columns even if the data dict want them to be required?

hadley:
Oh that's a thing? 😆

gaborcsardi:
The answer is not so straightforward, actually, because parquet tools including arrow create OPTIONAL columns by default. So I expect a lot of parquet files in the wild where columns are OPTIONAL even though they are never missing and they never meant to be missing.

hadley:
Ok, we should do a quick first check for REQUIRED then

PR #61, opened 2026-06-18
Add standard schema metadata

### LINKED ISSUE #4: Comment at top of file
Should give pointer to spec and or skills? Useful for both humans and agents to understand, and serves as basic understanding.

e.g. 

```yaml
# Learn more about data-dict.yaml at https://github.com/hadley/data-dict.yaml
```

Or possibly:

```yaml
$learn-more: https://github.com/hadley/data-dict.yaml
```

Advantage is that non-comments are more likely to get round tripped by all tools. `$` prefix is a common convention for "meta" fields. Would be most compelling if we used a spec:

```yaml
$schema: https://hadley.github.io/data-dict/schema.json
$learn-more: https://github.com/hadley/data-dict.yaml
$version: 1.0
```

hadley:
* `version` becomes `$version`
* Add `$learn-more`

Fixes #4

### DIFF
--- a/.claude/claude.md
+++ b/.claude/claude.md
@@ -64,7 +64,9 @@ YAML file
   → Result<(), Vec<Diagnostic>>
 ```
 
-### Lint rules (DD001–DD008)
+### Lint rules (DD001–DD009)
+
+DD001–DD008 are errors (they fail validation); DD009 is a warning (reported but does not fail validation).
 
 | Rule | Description |
 |------|-------------|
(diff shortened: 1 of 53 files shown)


.claude/claude.md
```diff
@@ -76,9 +78,19 @@ YAML file
 | DD006 | Cardinality inconsistent with column constraints |
 | DD007 | Column missing required representation key (`values`, `range`, or `examples`) |
 | DD008 | Column has `units` but its type is not `number(quantity)` |
+| DD009 | Document omits the recommended `$learn_more` key (warning) |
```
gaborcsardi, line 85: Feels like introducing warning is a big difference, are you sure about it?

hadley: Yeah, I think we will have more warnings going forward. 

We should have a principle for errors vs warnings, and I think it's "errors stop work; you must fix immediately"; warnings are something you should fix when you're next working on the project.

gaborcsardi:
Currently the warnings are not included in the JSON results, and they probably should.

hadley:
I have maybe gone overboard in refactoring here, but I think it's a nicer design

gaborcsardi:
No, the opposite, refactoring is good, especially at this stage of the project. I always meant the initial design as something simple that we'll refactor as we add complexity.

gaborcsardi (approved):

PR #85, opened 2026-06-24
VS Code extension PoC

gaborcsardi:
- Live validation of `data-dict.yaml` files. (Or you can configure a glob to live validate.)
- Data validation command, call it from the command palette.

We'll need to improve some of the error locations, because e.g. a missing top-level tag currently marks the whole document, which is ok for a cli error, but does not look good in the live validation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -139,10 +139,38 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d902e3d592a523def97af8f317b08ce16b7ab854c1985a0c671e6f15cebc236"
 
 [[package]]
+name = "async-trait"
+version = "0.1.89"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "auto_impl"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


hadley:
Would you mind updating this against main? I did a bunch of work to expose better context, and the individual problems now expose a location that you should be able to use.

gaborcsardi:
> Would you mind updating this against main? I did a bunch of work to expose better context, and the individual problems now expose a location that you should be able to use.

Done.

gaborcsardi:
You'll probably want to merge these for easier testing:
https://github.com/hadley/dabstep/pull/1
https://github.com/hadley/elevators/pull/1
https://github.com/hadley/foodbank/pull/1
https://github.com/hadley/loan-application/pull/1
https://github.com/hadley/otters/pull/2

hadley:
Now merged those, and updated the examples here. Thanks!

hadley:
This looks good! I'd suggest we merge and then iterate on the errors:

* I think https://github.com/posit-dev/quarto-yaml needs some tweaks to better report locations when something is missing
* There's something weird going on with strings? Like the span is ignoring the quotes
* We need to tweak the rendering of "actual" in these hints as the messages currently rely on context that's not shown, i.e. the second line isn't right here:

Screenshot 2026-07-07 at 10 15 20

Then can you take a look at autocomplete? Maybe start with `type`? It's probably not super important since I imagine these files will mostly be created by AI (with help from the cli) but if it's not too hard to add it would  be nice.

Can we add a button somewhere to make the data validation more obvious?

hadley:
We should also bundle the cli in the extension like air, quarto etc, https://github.com/posit-dev/air/pull/504, since I think that's a useful distribution mechanism.

PR #90, opened 2026-06-29
Add a data version

### LINKED ISSUE #14: Versioning?
When was it created/updated? etag/last-modified/version ??

hadley:
Fixes #14

@lwjohnst86 would you mind taking a look, primarily at the changes to `schema.md`? Anything you think is missing?

### DIFF
--- a/site/spec.md
+++ b/site/spec.md
@@ -215,11 +215,13 @@ glossary:
 
 `version` is optional. It's a map with exactly one of three keys, which names both the kind of version and its value:
 
-* `number`: a hand-curated version number, such as `1.2.0`.
+* `number`: a hand-curated version number with three dot-separated numeric components, optionally followed by a pre-release (`-…`) and/or build (`+…`) suffix, such as `1.2.0` or `1.2.0-rc.1`.
 * `date`: a release date in ISO 8601 form (`YYYY-MM-DD`), such as `2024-01-31`, for data refreshed on a schedule.
 * `hash`: an opaque identifier, such as `a1b2c3d`, derived from the data itself.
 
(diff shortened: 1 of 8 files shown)


lwjohnst86 (commented):
Perfect 👍 A very good addition 👍

site/spec.md
```diff
@@ -171,3 +170,20 @@ glossary:
...
     primarily by chemical analysis.
 ```
+
+## Version
+
+`version` records the version of the data this dictionary describes, so people and tools can tell two snapshots of the data apart and know which one a given dictionary goes with. (This is distinct from `$version`, which records the version of the *spec* the document conforms to.)
+
+`version` is optional. It's a map with exactly one of three keys, which names both the kind of version and its value:
+
+* `number`: a hand-curated version number, such as `1.2.0`.
```
lwjohnst86: Not sure if this is exactly the place for it, but could the check confirm whether it follows SemVer? Rather than just a number or string? Data Package has a nice post on how SemVer would work with data https://datapackage.org/recipes/data-package-version/ which was quite enlightening to reframe how we thought of treating data from a release point of view.

gaborcsardi:
Maybe we could allow a time stamp, not just a date, for data that changes more often? 

As for requiring semver, I would not go as far as requiring it, but we could nevertheless encourage it.

gaborcsardi (approved):

site/spec.md
hadley: Yeah, I think we can recommend semver, but I don't see any way to enforce it.

hadley: Lets at least enforce a three component version number.

PR #101, opened 2026-07-01
Standardise table naming

### LINKED ISSUE #97: Inconsistent naming for tables
Currently tables use a named key, but everything else uses `name: foo`. This seems like potentially confusing.

I do think we want to standardise of `name: foo` because for columns I was eventually imagining an syntax for describing multiple columns at once like `matches: ^foo` (which would be mutually exclusive with name). I don't _think_ we want something similar for tables, since I think we're most likely to handle at the source level with a glob, creating a single virtual table.

I think it's ok to leave `glossary` as is since, that's really "term" rather than "name".

hadley:
Now matches dataset and column, i.e. you use an explicit `name: foo`.

Fixes #97

@gaborcsardi does this make sense to you?


gaborcsardi (approved):
Definitely an improvement!

Issue #39, opened 2026-06-16
Extensions to checking metadata against the standard


lwjohnst86:
One thing we encountered when we first started using the Data Package standard is that their "MUST" requirements for what should be in a data package were very loose. It was really only one item: a "resources" field. But we wanted something that was a bit more opinionated. Like, it should have a contributor list and it should have a title and version, etc. But there was no mechanism in place to allow for that in the standard (there is an "extensions" section, but it's very difficult to understand and learn how to create something like that). So we ended up making our own checker that included extensions, e.g. a "strict" check or custom checks (https://check-datapackage.seedcase-project.org). Ideally there should be some way that any standard/spec have to include extensions to it and to modify what is required and not required.

hadley:
Since we're giving every check a unique id, I think this would be pretty easy to add. In the short term, the easiest way to add additional checks would be to include some stricter checks that are turned off by default, and then you could deliberately turn them on with your config.

On the flip side, there may also be default checks that you want to turn off. For example, checking for uniqueness (and I suspect foreign key matches) is going to be expensive, and if that's already guaranteed some other way, you might want to suppress that check.

I think eventually we could maybe export some of the current internals to make it possible for you to use data-dict as a rust library to create your own CLI (now achievable by scientists with the help of claude code), but I think that would be at least a year or two down the road as we'd want the internals to be very stable before exposing them to others.

hadley:
My running list of additional validations you might want to turn on:

* Validate range and examples against the data. Makes sense if it's a static dataset, but not if it's (e.g.) a daily export.
* Require a version
* Require labels for all columns
* Require a primary key for all tables

PR #111, opened 2026-07-09
Add `label`

### LINKED ISSUE #108: Consider a `label` field
e.g. so you can provide a longer human readable label for terse field names.

Add to name/description/details system.

hadley:
Fixes #108

@lwjohnst86 I'd love a second pair of eyes on this if you have the time 😄


crates/data-dict-cli/skills/write-data-dict.md
```diff
@@ -74,6 +74,9 @@ a plausible-sounding description for a column whose meaning you had to guess.
     c.  For each column, create an entry with:
 
         -   `name`: must match the actual column name exactly.
+        -   `label` (optional): a short, human-readable title, worth adding when
```
lwjohnst86, line 77: I've seen this called "title" in other formats, but "label" also works. "label" I feel is closest to how e.g. SAS uses the term. Either label or title works though. And yes, this is an important inclusion to have 😛

site/spec.md
```diff
@@ -9,11 +9,7 @@ The metadata keys are:
...
 
-The descriptive keys identify and document the dataset as a whole:
-
-* `name` (optional): a human-readable name for the dataset, suitable for display in a user interface that lists several dictionaries. Unlike a table name, it has no uniqueness or character constraints — it's a title, not an identifier.
-* `description` (optional): a short, human-readable description of the dataset. May contain markdown, and is usually a few sentences or a paragraph.
-* `details` (optional): additional information about the dataset. Can be any length.
+The descriptive keys — `name`, `label`, `description`, and `details` — identify and document the dataset as a whole. All four are optional here, and work the same way at every level of the dictionary; see [Name, label, description & details](#name-label-description--details) for their full meaning. For the dataset, `name` is a terse identifier (e.g. `foodbank`) and `label` its human-readable title.
```
lwjohnst86, line 12: I this isn't part of this PR, but I have a hard time know when I might use `description` vs `details`. I might make an issue about this though.

site/spec.md
```diff
@@ -107,9 +99,16 @@ Each column also needs describe some representative values, using exactly one of
...
+* `label` is a short, human-readable title, useful when the `name` is terse or technical (e.g. `FoodData Central ID` for `fdc_id`). Plain text (no markdown), typically a few words, it stands in for the `name` in user interfaces.
+* `description` is a short summary, typically a few sentences or at most a paragraph. May contain markdown, and is displayed in user interfaces. A good table description answers two questions — **what's the grain?** (what does a row represent, e.g. "each row is a food item") and **what's the population?** (what's been included or filtered out, e.g. "only completed orders from 2020 onwards").
+* `details` is a free-text note of any length: the place to carefully record everything else, such as "here be dragons" assumptions, known weak spots, surprising calculations, and how the data was collected or constructed.
```
lwjohnst86: I don't know what "here be dragons" means in this context. Maybe?

```suggestion
* `details` is a free-text note of any length: the place to carefully record everything else, such as assumptions about potential unknowns, known weak spots, surprising calculations, and how the data was collected or constructed.
```

site/spec.md
```diff
@@ -107,9 +99,16 @@ Each column also needs describe some representative values, using exactly one of
...
 
-The `description` and `details` are free text fields that humans and agents can use to jot down important notes. The `description` should be short, typically a few sentences or at most a paragraph and will be displayed in user interfaces. The `details` can be any length, and is a good place to carefully record all the details of the table.
+Every field but `name` is optional at every level.
```
lwjohnst86, line 111: Related to the "strictness" checks I mentioned in other PRs/issues, I think label should never be optional but understand that might put a big burden to pass the checks for many existing datasets. Could `label` be included as a "next level required" field? A bit like how Cargo clippy has different levels of linting?

hadley: I think the problem with making it required is that for many datasets you'll end with pretty redundant labels. But I made a note in #39 that this might be an optional check that folks want to turn on.

site/spec.md
hadley: Yeah, me too 😄

PR #113, opened 2026-07-09
Initial spec for lists and structs

### LINKED ISSUE #56: Consider list types

### LINKED ISSUE #57: Consider nested types

hadley:
Will eventually fix #56 and fix #57.


site/spec.md
```diff
@@ -136,6 +136,8 @@ The supported types are:
 * `date`: calendar dates, written as ISO 8601 strings (`YYYY-MM-DD`, e.g. `2024-01-31`).
 * `datetime`: date-times, written as ISO 8601 strings. Without a `time_zone` they carry an offset (e.g. `2024-01-31T09:30:00Z`); with a `time_zone` they're written zoneless and interpreted in that zone (see [Time zones](#time-zones)).
 * `enum`: a column with repeated values from a known set. The allowed values are listed in the `values` property.
+* `list(element_type)`: an ordered sequence of zero or more elements of the given type (see [List element types](#list-element-types)).
```
hadley, line 160: Are list and struct the right names?

site/spec.md
```diff
@@ -193,8 +248,8 @@ NB: when `time_zone` is present, write the column's `range` as plain, zoneless d
 
 The `constraints` property is a list of constraint names. The supported constraints are:
 
-* `primary_key`: the set of columns with the `primary_key` constraint uniquely identifies each row. Implies `required` and `unique`.
-* `foreign_key`: the column references a primary key in another table (or in the current table, if a self-join). The specific relationship is defined in [`relationships`](#relationships).
+* `primary_key`: the set of columns with the `primary_key` constraint uniquely identifies each row. Implies `required` and `unique`. Not valid on `list` or `struct` columns, or on fields within a `struct`.
```
hadley: We could consider consider other carve outs if you think support is going to be complex

site/spec.md
```diff
@@ -156,9 +158,60 @@ A `number(quantity)` column can also declare its `units`: a free-text string nam
...
+The element type in `list(element_type)` may be any type: `string`, `number`, `number(id)`, `number(ordinal)`, `number(quantity)`, `boolean`, `date`, `datetime`, `enum`, or `struct`. The same properties that apply to a column of that type apply when it is used as a list element type — `values` for `enum`, `fields` for `struct`, and so on.
+
+```yaml
+- name: tags
+  type: list(string)
+  examples: [nature, outdoor, urban, photography, wildlife]
+
+- name: categories
+  type: list(enum)
+  values: [food, drink, dessert]
```
lwjohnst86, line 193: This looks the same as an enum. What's different here?

site/spec.md
```diff
@@ -156,9 +158,60 @@ A `number(quantity)` column can also declare its `units`: a free-text string nam
...
 ```
 
+#### List element types
+
+The element type in `list(element_type)` may be any type: `string`, `number`, `number(id)`, `number(ordinal)`, `number(quantity)`, `boolean`, `date`, `datetime`, `enum`, or `struct`. The same properties that apply to a column of that type apply when it is used as a list element type — `values` for `enum`, `fields` for `struct`, and so on.
+
+```yaml
+- name: tags
+  type: list(string)
+  examples: [nature, outdoor, urban, photography, wildlife]
```
lwjohnst86, line 189: Same comment below about the enum. How would this be different?

site/spec.md
```diff
@@ -156,9 +158,60 @@ A `number(quantity)` column can also declare its `units`: a free-text string nam
...
+      examples: [123 Main St, 456 Oak Ave, 789 Elm Dr]
+    - name: city
+      type: string
+      examples: [Portland, Austin, Chicago]
+    - name: zip
+      type: string
+      examples: ["97201", "78701", "60601"]
+    - name: country
+      type: enum
+      values: [US, CA, MX]
```
lwjohnst86, line 237: I have a hard time seeing the difference here with the `list(struct)`. I understand that the data itself might be different, but when the metadata looks the exact same, I'd have a hard time differentiating them both.

site/spec.md
```diff
@@ -136,6 +136,8 @@ The supported types are:
 * `date`: calendar dates, written as ISO 8601 strings (`YYYY-MM-DD`, e.g. `2024-01-31`).
 * `datetime`: date-times, written as ISO 8601 strings. Without a `time_zone` they carry an offset (e.g. `2024-01-31T09:30:00Z`); with a `time_zone` they're written zoneless and interpreted in that zone (see [Time zones](#time-zones)).
 * `enum`: a column with repeated values from a known set. The allowed values are listed in the `values` property.
+* `list(element_type)`: an ordered sequence of zero or more elements of the given type (see [List element types](#list-element-types)).
+* `struct`: a structured record with named fields documented in the required `fields` property (see [Struct fields](#struct-fields)).
```
lwjohnst86, line 161: Are you thinking of using these for very nested data? I personally haven't encountered these "in the wild". What would be the advantage of a `list(struct)` rather than as either another table or as a flattened set of columns with specific items appended to their column name.

site/spec.md
```diff
@@ -156,9 +158,60 @@ A `number(quantity)` column can also declare its `units`: a free-text string nam
...
+      range: [1, 100]
+    - name: price
+      type: number(quantity)
+      units: USD
+      range: [0.99, 999.99]
+```
+
+#### Struct fields
+
+A `struct` column may include a `fields` property — an ordered list of field descriptors. Each field descriptor uses the same schema as a column descriptor. A field may itself be `list(...)` or `struct` (with its own `fields`), allowing deep nesting.
```
gaborcsardi: So if I understand correctly `list(..., struct(list(...)), ...)` is allowed, but `list(list(...))` is not? Is that intentional? Since the list -> struct -> list nesting is already allowed I don't see why the `list(list(...))` would not be.

site/spec.md
gaborcsardi: list definitely. struct is as good as any imo

site/spec.md
gaborcsardi: This looks OK for me, although I am not sure what more you would add.

gaborcsardi (approved):

site/spec.md
hadley: It's a multiselect — an `enum` must have exactly one value for each row. A `list(enum)` can have any number of values for each row.

site/spec.md
hadley: Ah yes, good point.

site/spec.md
hadley: It's not a data structure that I'd generally recommend, since I would prefer a relation framing with multiple tables, but it is something that you get in parquet files so its important that we can represent it.

hadley:
To do:

* [x] Carefully review snapshots
* [x] Check that list/struct/list recursion is correctly implemented
* [x] Support list-list nesting

PR #121, opened 2026-07-15
Rework the "why" section in the readme

### LINKED ISSUE #115: Include more human in the "why data-dict"
Related to my issue #114, when I first read the "tagline" for data-dict, I saw it as something I as a data producer could use and benefit from. But then I got to the "why data-dict" and I didn't see me or what I/we do there. So I felt a disconnect, was this actually for me?

I think a big thing would be to include more human in the "why" rather than AI. I know you're very optimistic about AI, but as a data producer, I work in environments where 1) we're creating the data where there is no context, no documentation, so AI isn't very useful and 2) our servers don't allow AI in them because of legal and privacy reasons. I know in the US, laws are less restrictive about how human (especially health) data is used, but in Europe and with GDPR, using any AI in these settings is difficult if not impossible, depending on the legal team of the institution. So we don't have the same access to these tools, nor do we see the same benefit to using them in these settings. I think it would help a lot to include what value data-dict has for the human data producers. I would also say that data documentation has *always* been about helping humans and it shouldn't be because of AI that this is important to do now.

hadley:
Closes #118. Closes #117. Fixes #115.

### DIFF
--- a/site/index.md
+++ b/site/index.md
@@ -2,7 +2,7 @@
 title: "data-dict.yaml"
 ---
 
-`data-dict.yaml` is a data dictionary specification that describes a collection of related tables: their contents, constraints, connections, and the specialised vocabulary you need to understand them. It is designed to be a living document, co-written by humans and agents, that tracks your understanding of a dataset as it evolves.
+`data-dict.yaml` is a data dictionary specification that describes a collection of related tables: their contents, constraints, connections, and the specialised vocabulary you need to understand them. It is designed to be a living document, accessible to both humans and agents, that tracks your understanding of a dataset as it evolves.
 
 `data-dict.yaml` is designed to be lightweight. It doesn't attempt to precisely describe every possible type of metadata in a machine-readable way. Instead it focuses on precisely recording the most important components, leaving the remainder to plain text fields that require a human or agent to interpret. This means that `data-dict.yaml` doesn't itself do **data cleaning**, but it is a useful complement to tools that do.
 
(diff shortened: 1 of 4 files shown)


site/index.md
```diff
@@ -14,24 +14,46 @@ You can read the details of the spec in [the specification](spec.md), or dive in
...
-* Unlike previous data dictionaries, we assume data is stored in parquet files or database tables. This means that many parsing details are out of scope, radically simplifying the spec.
-* The cost of describing the data semantics in multiple places (i.e. `data-dict.yaml` and data transformation code) is lower because an AI agent can easily keep both in sync.
+The specification is:
+
+* Open and supported by Posit, a public benefit corporation with a mission to create free and open-source software for data science, scientific research, and technical communication. We don't yet have a formal governance model, but in the meantime, you're very welcome to propose changes and additions in the [issues](https://github.com/tidyverse/data-dict/issues).
+
+* Built by data scientists, for data scientists. We understand the challenges data scientists face, and we've designed the spec to address them directly. This means including pieces that other dictionaries omit: support for metadata like data version and column units, descriptions of the relationships between datasets, and a glossary for domain- and team-specific terminology.
```
lwjohnst86: Very nice improvements from my PRs 🥳 now you've stated your target user, so just to be clear, you aren't designing this for or consider how/if data engineers, research software engineers, data stewards and data managers would use it?

hadley: Hmmmmmmmmmmmmm, that's a good point. I think I've probably overweighted data scientists because I love that phrase so much.

hadley: And thanks for the feedback and other PRs. It's super helpful and I really appreciate it.

lwjohnst86: The change is much better and more inclusive, thanks! :grin:

PR #133, opened 2026-07-22
Add origin field linking a dataset or table to its source code

### LINKED ISSUE #13: Link to source code
If dataset was produced by a script, nice to be able to link to it.

hadley:
Adds `origin`, an optional field that links a dataset — or an individual table — to the code that produced it, such as a data-prep script or pipeline, giving a reader (human or agent) a starting point for understanding how the data was built. The value is a loose reference: either a URL or a path resolved relative to the dictionary file. It's accepted at the dataset (top) level and on each table. Fixes #13.

## Example

```yaml
name: foodbank
origin: https://github.com/hadley/foodbank
tables:
  - name: food
    origin: data-raw/food.R
    columns:
      - name: id
        type: number(id)
        examples: [1, 2, 3]
```

Use the dataset level when one pipeline produces everything, and the table level when different tables are built by different scripts.

## Design decisions

The issue suggested calling this `source`, but `source` is already the table-level key describing how to *read* the data, so reusing it would collide. `origin` keeps the two ideas separate: `source` is where the data lives, `origin` is where the code that made it lives.

`origin` is deliberately unenforced — there's no validation check, and the validator never fetches a URL or checks that a path exists. It's a loose pointer for a human or agent to follow, and a path can't be reliably told apart from arbitrary text, so there's nothing meaningful to validate. It stays a bare optional string in the schema, like `$learn_more`.

### DIFF
--- a/crates/data-dict/tests/validate_spec.rs
+++ b/crates/data-dict/tests/validate_spec.rs
@@ -220,6 +220,38 @@ fn top_level_description_no_s16() {
     "});
 }
 
+// `origin` is a loose, unenforced reference (a URL or a dictionary-relative
+// path) accepted at both the dataset and table levels.
+#[test]
+fn origin_dataset_and_table() {
+    assert_clean_dict(indoc! {"
+        name: foodbank
+        origin: https://github.com/example/foodbank/blob/main/data-raw/all.R
+        tables:
+          - name: food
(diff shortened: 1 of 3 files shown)


gaborcsardi (approved):
Makes sense.

Issue #9, opened 2026-04-22
Think about labelled data more


hadley:
i.e. https://haven.tidyverse.org/reference/labelled.html

In particular, we need some way to label numbers, and some way to mark multiple values as missing.

This would really help folks in the social sciences and those who use spss/stata/SAS.

nathanj3:
My two cents (as one of the social sciences folks who attended the Data Science Lab on Tuesday and uses `labelled` and SPSS regularly):

## Missing values

I'm mentioning this one first, since missing values are sometimes also labelled, but the missingness is a property of the value (not the label).

The `columns` descriptor could add a `missing` property. Something like:

- **missing**: an optional list of missing values.
    - for any column type, a list specifying each possible missing value, e.g. `missing: NA`, `missing: ""`, or `missing: ["No response", "Skipped"]`
    - for an `enum` or `number` column, an alternative is to supply one or more of the following, possibly in addition to the above:
        - a prefix for all possible missing values, e.g., `missing(prefix): .` would treat `.` and `.A` as missing; `missing(prefix): NA` would treat `NA` and `NA(a)` as missing
        - a two-element list by using `missing(range): [min, max]` to indicate a range of missing values (using `Inf` and `-Inf` for $\pm \infty$)

In this case, SAS- and Stata-style missings could be handled as `missing(prefix): .` (or `missing(prefix): NA` if read into R by {haven}). SPSS style missings could be handled as:

```
missing: 
  - range: [95, Inf]
  - -1
```


## Labelled numbers

For `enum`, if the values are a map, the numbers would be the values, and it's easy to add optional `missing`s:

```
- name: workshop_rating
  description: Please rate the quality of the workshop you attended.
  type: enum
  values:
    1: Very poor
    2: Poor
    3: Fair
    4: Good
    5: Excellent
    98: Did not attend the workshop
    99: No response
  missing: [98, 99]     # OR: missing(range): [98, Inf]
````

For `number` columns, adding an optional `values` (or `labels`?) property that maps the labels to the values becomes very similar to `enum`. My initial instinct was to just use `enum` for labelled numbers -- but I'm thinking about a few cases in particular where `number` may make more sense:

**Dichotomous data**

It's easy enough to encode dichotomous/dummy-coded columns using `enum`, but it may be useful to specify in some way that averaging and summing are valid operations on this column. Making an `enum(quantity)` type would be an option, but `number(quantity)` already exists.

```
- name: attended_workshop
[...]

PR #139, opened 2026-07-27
Document expression language in more detail

hadley:


hadley:
Pushed d97855b, which also makes some changes to the type ulres:

**Enums now take the type of their `values`.** A string enum is a string and a number enum a number, both in the data and in expressions, so `LENGTH(sex) = 1` is checked and `LENGTH(grade) = 1` on a number enum is now S21. Previously an enum matched anything. I _think_ this is what we want but does imply that an enum is just a restricted set of values from a regular type, not a class enum.

**Untyped columns are an error where their type matters (S23).** A column listed by name only used to be a wildcard; now it's an error. Operands whose type is never consulted are exempt, so `notes IS NOT NULL` and `COLUMNS(*) IS NOT NULL` on a table with undocumented columns stay legal.

Internally `Ty` gained `Unknown` as the opposite of `Any` — compatible with nothing rather than everything. One incidental fix came with it: a node visited twice (inferred by its parent, then required by it) reported the same fault twice, so `missing + 1 > 0` gave two S20s; findings are now deduped.

Also filed #141 — the schema still permits `values: [1, unknown]`, and a mixed enum now quietly loses type checking, so it's worth rejecting outright.

gaborcsardi:
Re enums, IDK if we need to allow an enum on any (simple) type. It would be definitely simpler to only allow it on strings, like in Parquet: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#enum

Also, currently you don't need to define the type of the enum. So e.g. for
```
      - name: foo
        type: enum
        values: [1,2,3]
```
what should be the type of the Parquet column? Integer? Or string is also ok?

hadley:
Ah, yeah, that's a good point. I was getting confused with https://github.com/tidyverse/data-dict/issues/9, but that's a slightly different feature because you need to be able to (e.g.) label some integers, not all of them. Let me make enums always strings in the other PR, and then circle back to this one.

site/expressions.md
```diff
@@ -0,0 +1,363 @@
...
+
+## Functions
+
+Function names are case-insensitive. Every function is null-propagating: any null argument gives a null result. Passing the wrong number of arguments, or an argument of the wrong type, is an error at spec-validation time, not something that happens per row.
+
+### String functions
+
+#### `LENGTH(s)`
+
+`string → number`. The number of characters (Unicode code points) in `s`, not the number of bytes.
```
gaborcsardi, line 190: I wish this was number of graphemes and not code points. But apparently in SQL it is usually code points, so it makes sense to keep it as code points...

gaborcsardi (approved):
Looks great! Yeah, we very much needed this.

I added a commit to express the decision that ENUMs are always strings: [`3e5cb3a` (this PR)](https://github.com/tidyverse/data-dict/pull/139/changes/3e5cb3a1fc228430762a6a1798f57e1f76761f24)

Btw. do we want to have some way to use column names that clash with reserved words in expressions? Right now there is no documented way, although there is an undocumented accidental way with `COLUMNS()` that sometimes works:
```
COLUMNS([BETWEEN]) > 0
```
We could support backticks or double quotes for this. Or document `COLUMNS()`, although that only works for a single such column in an expression, because multiple `COLUMNS()` are not allowed in a single expression.

hadley:
Yeah, we should specify a quoting syntax. I don't love using `"`, like SQL, but I guess that's standard and is what duckdb uses. But I could be convinced that `` ` `` is better.

gaborcsardi:
Well, BigQuery, SQLite and MySQL use or at least accept `` ` ``, but everything else is double quotes, including DuckDB.  (Advanced) R uses will like the backtick, tough. 

Btw. this is not only about reserved words because columns may include special characters, but identifier cannot, so any column name that is not just letters + numbers + underscores is currently not allowed in constraints.

gaborcsardi:
FWIW, I don't like the double quotes, either, and some SQLs already use backticks, so I think it is OK if we use backticks. Double quotes are syntax error currently, which helps avoiding confusions.

We could also allow both, SQLite does that I think.

hadley:
I'd rather stick with one syntax and backticks are fine with me. I think double quotes are also likely to be confusing to python users.

gaborcsardi:
Sounds good! I don't think this blocks this PR, btw. I think we can add the backticks in another PR.

Issue #147, opened 2026-07-29
`data-dict draft`: generate a skeleton data-dict.yaml from parquet files


hadley:
```
data-dict draft <PATHS>...  [-o <FILE>]

  -o, --output <FILE>   where to write [default: ./data-dict.yaml]; `-` for stdout
```

Generates a starting dictionary from one or more parquet files, built on the profiling engine (#145).

- **New file**: a complete, `validate-spec`-clean dictionary: `$version`, `$learn_more`, one table per file (`name` = file stem, `source.parquet` = path relative to the output file), inferred columns.
- **Existing file**: parse it, then **append** tables only for files whose table name isn't already present, inserting after the last entry in the `tables` list (using tracked spans) and leaving existing text byte-for-byte untouched.
- Two inputs with the same stem → error before writing anything.
- Everything the human/llm must review goes in as `# TODO:` YAML comments, so the output always passes `validate-spec`.

Inference rules:

| From | Inferred |
|---|---|
| Parquet logical/physical type | `type`, via `parquet_type_to_dict_type` |
| Column named `id` / `*_id` with a number type | `number(id)` |
| Timestamp `isAdjustedToUTC` | `time_zone: UTC` (true) / `time_zone: naive` (false) |
| `date` / `datetime` | `range: [min, max]` |
| `number` / `string` | `examples:` — 5 evenly spaced along sorted uniques |
| Zero nulls observed | `# TODO:` comment suggesting `constraints: [required]` |
| String with ≤ 12 distinct values, clearly repeated (distinct < rows/2) | `# TODO:` comment suggesting `type: enum` with the observed values |
| Footer proves distinct = rows | `# TODO:` comment suggesting `unique` / `primary_key` |

Example output:

```yaml
$version: 0.1.0
$learn_more: https://data-dict.tidyverse.org

# TODO: describe the dataset as a whole.
# name:
# label:
# description:

tables:
  - name: penguins
    source:
      parquet: data/penguins.parquet
    # TODO: what's the grain? what's the population?
    # description:
    columns:
      - name: species
        type: string
        examples: [Adelie, Chinstrap, Gentoo]
        # TODO: only 3 distinct values, consider:
        # type: enum
        # values: [Adelie, Chinstrap, Gentoo]
      - name: bill_length_mm
        type: number
        examples: [32.1, 39.5, 44.1, 48.6, 59.6]
        # TODO: no missing values observed, consider constraints: [required]
```

[...]

Issue #146, opened 2026-07-29
`data-dict describe`: summarise the columns of a parquet file


hadley:
```
data-dict describe <PATH> [COLUMN]  [--json]
```

Human-readable and machine-readable per-column summaries of a parquet file, built on the profiling engine (#145). All columns by default; positional `COLUMN` narrows to one. Dispatches on file extension (`.parquet` only for now; clear error otherwise). `--json` emits the raw profiles, matching the `validate-*` flag.

Per column: dict type + parquet type, **distinct** count, then **missing** (in that order — missing also appears as a histogram bin), then a type-appropriate body:

- **numeric / date / datetime** — min–max plus a unicode histogram, with a final `missing` bar (and a `NaN` bar for float columns when present) rendered from the counts the engine keeps outside the bins.
- **string / enum** — the 20 most common values with counts and bars, plus an `(n other values)` line.
- **boolean** — true/false/missing counts.

```
penguins.parquet: 344 rows × 8 columns

bill_length_mm — number (DOUBLE)
  distinct   164
  missing    2 (0.6%)
  32.1 – 34.9  ▇▇▇▇          13
  34.9 – 37.6  ▇▇▇▇▇▇▇▇▇▇▇▇  56
  ...
  57.0 – 59.6  ▇              3
  missing      ▏              2

species — string (BYTE_ARRAY / String)
  distinct   3
  missing    0
  Adelie     ▇▇▇▇▇▇▇▇▇▇▇▇  152
  Gentoo     ▇▇▇▇▇▇▇▇▇     124
  Chinstrap  ▇▇▇▇▇          68
```

`--json` serialises the profiles directly (serde on the engine's types, as `ProblemSet` does for the validators). Scalars use their YAML-compatible forms: numbers and booleans as themselves, dates/datetimes as ISO 8601 strings. Keys that don't apply to a column's type (`min`, `max`, `histogram`) are omitted. There's no separate "approximate" flag — `distinct: {"approx": ...}` is the signal. `[COLUMN]` narrows the `columns` array the same way it narrows the text output.

```json
{
  "path": "penguins.parquet",
  "rows": 344,
  "columns": [
    {
      "name": "bill_length_mm",
      "type": "number",
      "parquet_type": "DOUBLE",
      "distinct": { "exact": 164 },
      "missing": 2,
      "min": 32.1,
      "max": 59.6,
      "histogram": {
        "bins": [
          { "lower": 32.1, "upper": 33.5, "count": 4 },
[...]

PR #149, opened 2026-07-29
Add a column profiling engine (#145)

### LINKED ISSUE #145: Column profiling engine in `data-dict-parquet`
Add a `profile.rs` module to `data-dict-parquet` that computes per-column summaries of a parquet file. This is the shared engine behind the upcoming `describe` (#146) and `draft` (#147) commands, which are just two views over it.

```rust
pub struct ColumnProfile {
    pub row_count: usize,
    pub null_count: usize,
    pub distinct: Distinct,        // Exact(n) | Approx(n) once the tracked-value cap is hit
    pub min: Option<Value>,        // ordered types only
    pub max: Option<Value>,
    pub value_counts: Vec<(Value, usize)>,  // top-k, k = 20
    pub histogram: Option<Histogram>,       // numeric/temporal
    pub examples: Vec<Value>,      // 5 evenly spaced along sorted uniques
}

pub struct Histogram {
    /// 20 equal-width bins spanning [min, max]. Nulls are never counted
    /// here — that's `null_count`, which applies to every type.
    pub bins: Vec<Bin>,
    /// NaN values, for float columns only (NaN is neither null nor orderable,
    /// so it gets its own bin).
    pub nan_count: usize,
}

/// Counts values in (lower, upper], or [lower, upper] when `lower_inclusive`.
/// Only the first bin is lower-inclusive, so the minimum has a home — but the
/// flag is carried on the bin, not left as a convention, so every consumer
/// (renderer, JSON) sees the boundary rule explicitly.
pub struct Bin {
    pub lower: Value,
    pub upper: Value,
    pub lower_inclusive: bool,
    pub count: usize,
}

/// Per-column accumulator for distinct values and their counts.
struct ValueCounts {
    /// Occurrence counts for up to 1,000 tracked values. Once full,
    /// already-tracked values keep counting (their totals stay exact), but
    /// unseen values are no longer added.
    counts: HashMap<Value, usize>,
    /// Set when a new value arrives after `counts` is full, i.e. at least one
    /// distinct value went untracked.
    saturated: bool,
    /// Every value, tracked or not, is fed to this sketch; it answers the
    /// distinct count once the map saturates.
    sketch: HyperLogLog,
}
```

`distinct`, `value_counts`, and `examples` all derive from `ValueCounts`:

- unsaturated → `Distinct::Exact(counts.len())`, exact top-20, exact examples.
- saturated → `Distinct::Approx(sketch.estimate())`. The 1k cap only bounds the *tracked* values; the HyperLogLog sketch counts distinct values without bound in ~16 KB per column (precision 14, ~1% typical error). The top-20 and examples come from the tracked values only — tracked counts are still exact totals, but a value first seen *after* saturation is invisible, so a late-arriving heavy hitter can be missed — both are marked approximate.
- HLL is ~100 lines to hand-roll; per the benchmarking guidance, measure a candidate crate (binary size, deps, build time) against hand-rolling before taking the dependency.

The dictionary fast path feeds the same structure (counting dictionary indices, then resolving through the page), so the cap and saturation semantics are identical on both paths.

Design:

- **Single streaming pass** per file with per-column accumulators, following the existing `column_stats` pattern. Accept an optional column filter so `describe <file> <column>` only reads one column.
- **Dictionary-page fast path** (as in D03/uniqueness): dict-encoded columns get exact distinct counts and value counts from dictionary pages + index counting, without decoding values.
- **Fallback** for plain-encoded columns: decode values into the same `ValueCounts` accumulator, with the cap and saturation semantics above.
- **Histogram**: 20 bins between min and max taken from footer statistics so bins are known before the pass; fall back to a second pass when stats are absent. Nulls are never binned (`null_count` covers every type); NaNs in float columns get their own bin.
- **Examples**: the spec's recommended baseline — 5 evenly spaced values along the sorted unique values, drawn from the same capped (≤ 1k) distinct set.
- Expose whatever type metadata downstream needs that `column_type_info` doesn't already carry (notably timestamp `isAdjustedToUTC`, for `time_zone` inference in `draft`).

Tests: generated parquet fixtures covering dict-encoded vs plain columns, multi row-group files, files without footer statistics, and every dict type (number, string, boolean, date, datetime).

hadley:
Closes #145.

Adds a `profile` module to `data-dict-parquet` that summarizes each column of a parquet file in one streaming pass and fixed amount of memory: row and null counts, distinct count, extremes, most frequent values, a histogram, and a few examples. This is the engine only — the `describe` (#146) and `draft` (#147) commands are two views over it. Nothing outside the crate changes except its `lib.rs` re-exports, and there are no spec or schema changes, since this adds no format or validation-check surface.

## Design

```rust
pub fn profile(path: &Path, columns: Option<&[&str]>) -> Result<Vec<ColumnProfile>, ParquetError>
```

Every summary is bounded, so a column of a billion distinct values costs the same memory as one with ten, and reaching a bound marks the affected numbers approximate rather than making them silently wrong. Following the review on the issue:

- **Frequent values** use a Space-Saving map (exact below 1,000 distinct values, then evict-min with the inherited count kept as a per-item error bound).
- **Examples** come from a bottom-k (KMV) sample at k=128, a uniform sample of the *distinct* set — a billion rows of `"US"` make it no likelier to be shown than a value occurring once. The reported five are spread evenly along the sorted sample, matching the baseline in `site/spec.md`.
- **Distinct counts** fall back to a precision-14 HyperLogLog once the tracker saturates.

Dictionary-encoded chunks are counted from their dictionary page plus the run lengths of the indices pointing into it, without decoding a value. The decision is per column chunk, into chunk-local counts, so bailing partway leaves nothing for the fallback scan to double-count.

## Types

Only types that can be counted, ordered and hashed are profiled. Decimals, UUIDs, fixed-length binary, INT96, half-floats, JSON/BSON and nested or repeated columns report row and null counts alone. Note this leaves a seam worth revisiting later: `types parquet` maps decimals to `number`, but a decimal column profiles to a stub.


crates/data-dict-parquet/src/profile.rs
```diff
@@ -0,0 +1,888 @@
...
+    meta.row_groups()
+        .iter()
+        .try_fold(0usize, |total, group| {
+            group
+                .column(leaf)
+                .statistics()
+                .and_then(|statistics| statistics.null_count_opt())
+                .map(|nulls| total + nulls as usize)
+        })
+        .unwrap_or(0)
```
gaborcsardi: This will report 0 missing values for row groups / column chunks that have no statistics, no?

crates/data-dict-parquet/src/rle.rs
```diff
@@ -0,0 +1,225 @@
+//! Decoder for Parquet's RLE / bit-packing hybrid encoding.
+//!
+//! Both the definition levels and the dictionary indices of a data page use it,
+//! and the profiler reads them straight from the raw page buffer. Values are
+//! reported a run at a time, so a page that repeats one dictionary index a
+//! million times costs one callback rather than a million.
+//!
+//! Parquet's own decoder is behind its `experimental` feature, which carries no
+//! semver guarantee, so this is hand-rolled against the (frozen) format.
```
gaborcsardi: I understand that this probably enables some performance optimizations, but writing our own RLE does not seems like a very good idea. Although it does not seem probable that parquet would add a new encoding for this, we cannot rule it out of course.

crates/data-dict-parquet/src/rle.rs
```diff
@@ -0,0 +1,225 @@
...
+    /// final group can run past the values the page declares; the padding is
+    /// skipped but its bytes are still consumed.
+    fn bit_packed(
+        &mut self,
+        groups: usize,
+        remaining: usize,
+        emit: &mut impl FnMut(u32, usize),
+    ) -> Result<usize, ParquetError> {
+        let width = self.bit_width as usize;
+        let data = self.take(groups * width)?;
```
gaborcsardi: Claude says that a crafted parquet file can make this overflow and cause a crash.

crates/data-dict-parquet/src/profile.rs
```diff
@@ -0,0 +1,888 @@
...
+/// its values; nothing is reported to `observer` in that case, so the fallback
+/// can't double count.
+fn count_dictionary(
+    row_group: &dyn RowGroupReader,
+    chunk: &ColumnChunkMetaData,
+    planned: &PlannedColumn,
+    repr: Repr,
+    observer: &mut impl Observe,
+) -> Result<bool, ParquetError> {
+    if chunk.dictionary_page_offset().is_none() || !encoding_stats_all_dictionary(chunk) {
```
gaborcsardi: Just because not all chunks are dictionary encoded we don't have to throw away the ones that are. For a writer that decides on whether to dictionary encode something or not by chunk, there might be a few chunks that are not dictionary-encoded, but most are.

crates/data-dict-parquet/src/profile.rs
```diff
@@ -0,0 +1,888 @@
...
+                .column(leaf)
+                .statistics()
+                .and_then(|statistics| statistics.null_count_opt())
+                .map(|nulls| total + nulls as usize)
+        })
+        .unwrap_or(0)
+}
+
+fn profile_column(path: &Path, target: &Target) -> Result<ColumnProfile, ParquetError> {
+    let file =
```
gaborcsardi: Ideally we would not re-parse the footer for every column.

gaborcsardi (approved):
It feels like this works on a lower level of Parquet than ideal, e.g. writing our RLE decoder. (E.g. does it work on big-endian platforms?) Probably not a huge risk, though. 

We should probably implement it for DECIMAL columns, because they are not uncommon (I think), and we do match that to `number` currently, so we already support it.

hadley:
Updated for the arrow port (#160): merged `main` and moved the profiler's value scan onto the arrow record-batch reader.

- **The dictionary-page fast path is untouched.** Counting a chunk from its dictionary page plus RLE index runs has no arrow equivalent (arrow doesn't expose raw pages), so `count_dictionary`, `rle.rs`, and `page.rs` still work the low-level API — only the fallback scan changed.
- **`scan_values` now reads arrow batches** per row group (`ValueColumn` in `value.rs` replaces the old `ColumnBatch`-based `batch_value`), reusing the `FileContext` reader from #160 with a new single-row-group projection.
- **Two small semantic upgrades fall out of arrow decoding:** narrow unsigned integers arrive as unsigned arrays, so the sign-extension masking now only applies on the raw dictionary path; and a parquet ENUM column (which arrow decodes as binary) profiles as text like any other string column.
- **One edge moved:** a column *declared* STRING whose bytes aren't UTF-8 now fails arrow's decode validation as a read error instead of downgrading to `Unsupported("non-UTF-8")`; the downgrade still applies to unannotated BYTE_ARRAY columns (the tested path — a declared-STRING file with invalid bytes is corrupt, and validate-data treats it the same way on main).
- Conflicts were confined to `lib.rs` (module list) and `dictionary.rs`, which keeps main's string-only enum semantics on top of this branch's shared page-walk helpers.

All 27 crate unit tests and the 22 profiling integration tests pass unchanged, along with the rest of the workspace.

hadley:
Follow-up in f8a7cd6: the hand-rolled RLE reader is gone — dictionary chunks are now read through the standard crate.

Instead of walking raw pages (page-header parsing, the RLE/bit-packed hybrid decoder, PLAIN dictionary decoding), the profiler asks arrow for the column as a `Dictionary(Int32, _)` array: the chunk's distinct values decode once, rows arrive as indices, and observing is a tally per distinct value per batch. Net −366 lines (`rle.rs` and the raw decode paths in `profile.rs`/`value.rs` deleted; `Repr` shrinks to the two pre-decode questions that remain: unprofilable types and untrustworthy unsigned footer stats).

One subtlety the footer forced: `page_encoding_stats` are optional, and a writer that hits the 1 MB dictionary-page limit falls back to plain encoding mid-chunk — requesting dictionary reads for those chunks made arrow hydrate every plain page into per-batch dictionaries (+31% on high-cardinality ints). A dictionary page near the limit is the fallback's fingerprint, so `prefer_dictionary` skips those; wrong guesses in either direction stay correct, they just cost time.

Measured (3M-row fixture, best of 3, old = RLE counting):

| Column | RLE counting | arrow dictionary |
|--------|-------------|------------------|
| 1000-distinct strings (dictionary) | 7 ms | 27 ms |
| 3M-distinct ints (fallback) | ~280 ms | ~272 ms |
| 3M-distinct strings (fallback) | ~570 ms | ~645 ms |

The low-cardinality case keeps its order-of-magnitude win over a plain read (~500 ms); the 20 ms it gives back is per-batch rather than per-chunk aggregation. The D04 fast path in `dictionary.rs` still reads dictionary *pages* (membership needs no index decoding, so it never touched the RLE reader), which is why `page.rs` survives.

hadley:
f183e43: float columns are no longer counted per value.

- `ColumnProfile::distinct` is now `Option<Distinct>` — `None` for `ValueKind::Float` (floating-point noise makes distinct counts misleading; the histogram and min/max carry the shape) and for never-read columns, where `Exact(0)` was overstating what we knew.
- `value_counts` is likewise empty for floats, since frequent-value counts lean on the same shaky equality.
- The sample of unique example values **stays** for floats: its bitwise dedup only exists to spread the sample, so a split pair of signed zeros there is harmless.
- With no per-value counting left, the `-0.0 → 0.0` canonicalization in `F64::new` is gone; `F64` equality now follows `total_cmp`, which also fixes a latent `Eq`/`Ord` disagreement about the zeros (`==` said equal, `total_cmp` said less).

hadley:
Good catch — and it was subtly worse than reported: the `try_fold` short-circuits on the *first* chunk lacking a null count, so one stats-less row group zeroed the reported count for the whole column, indistinguishable from a genuine zero.

Fixed in 85beb0d: `ColumnProfile::null_count` is now `Option<usize>` — always `Some` (exact) for scanned columns, the footer's sum for never-read columns, and `None` when any chunk omits the statistic, so an unknown count can't pose as zero. Pinned by `an_unread_column_without_statistics_has_no_null_count`.

Issue #137, opened 2026-07-25
Column/table names may contain characters that `join`/assertion expressions can't reference (no quoting)


jat255:
I came across this while playing around with generating data-dict.yaml schemas from LinkML examples. Not sure if this is intended design or perhaps just an early rough edge.

(the following was generated by Claude, but reviewed/edited by me)

## Summary

A table or column may be *declared* with a name containing spaces (or other characters outside `[A-Za-z0-9_]`, or a leading digit), but that name cannot be *referenced* from a relationship `join` expression or an assertion, and there's no quoting/escaping syntax to bridge the gap. The result is a column you can declare but can never form a relationship or assertion on.

## Details

Name declarations are validated only for non-emptiness. In `crates/data-dict/src/validate_spec.rs`, `validate_s11_table_name` and `validate_s11_column_name` (S11) reject a name only when it `.is_empty()` — there is no character-set restriction. So `name: creation date` is a valid column declaration and passes `validate-spec`.

But the `join` mini-language restricts identifiers to a strict grammar. From `crates/data-dict/src/join_expr.rs`:

```text
qcol  := IDENT "." IDENT
IDENT := [A-Za-z_][A-Za-z0-9_]*
```

`parse_ident` accepts only ASCII alphanumerics and `_`, and there is no production for a quoted/escaped identifier. The assertion mini-language (`crates/data-dict/src/assert_expr.rs`) uses the same `IDENT` rule; it has single-quoted *string literals*, but those are values, not a way to quote an *identifier*.

So a `relationships[].join` such as `Study.creation date = Other.id` fails to parse (S04), and there is no syntax — `"..."`, backticks, `[...]`, or backslash-escapes — that lets you name a space-containing column in a join.

## Reproduction

A spaced column name is a legal declaration — this validates clean:

```yaml
$version: 0.1.0
$learn_more: http://data-dict.tidyverse.org/
name: demo
tables:
  - name: A
    columns:
      - name: some field
        type: string
        examples: ["x"]
```

```
$ data-dict validate-spec ok.yaml
ok.yaml: ok
```

But that same name cannot be referenced from a join — this fails, with `S04` as the only error:

```yaml
$version: 0.1.0
$learn_more: http://data-dict.tidyverse.org/
name: demo
tables:
  - name: A
    columns:
      - name: id
        type: number(id)
        constraints: [primary_key]
        examples: [1]
  - name: B
    columns:
      - name: a ref
        type: number(id)
        examples: [1]
[...]

PR #155, opened 2026-07-31
Allow backtick-quoted names in expressions

### LINKED ISSUE #148: Define quoting in expression syntax
Using backticks

hadley:
A table or column could be declared with any non-empty name, but only a plain identifier (`[A-Za-z_][A-Za-z0-9_]*`) could be *referenced* from an assertion or a `join` — so a column called `creation date` was declarable and then unreachable, with no quoting syntax to bridge the gap.

Names of any other shape — containing a space or punctuation, starting with a digit, or colliding with a reserved word — now go in backticks:

```yaml
constraints:
  - assert: '`creation date` <= `end`'
relationships:
  - join: '`other studies`.`study id` = study.id'
```

A quoted name may hold any character, doubling a backtick to embed one. Each side of a join's `.` is quoted on its own, so a `.` between backticks is part of the name. Quoting changes only how a name is read, never how it's matched, so backticks are optional on a name that doesn't need them. Malformed quoting (unterminated, empty) lands on the existing S19/S04 rather than a new check code, so `validation.md` and `schema.yaml` are unchanged.

Backticks rather than SQL's `"` because `'…'` is already a string literal (and `COLUMNS('regex')` a regex), and because it matches the analytics mini-languages users come from — pandas `query()`, dplyr, Spark. One wrinkle documented in the spec: a backtick can't open a plain YAML scalar, so an expression *starting* with one needs YAML quotes; one merely containing a backtick doesn't.

Spec first: `site/expressions.md` gains a `Column references` section, plus grammar and `COLUMNS([...])` updates.

Tests: quoted names, reserved-word names, embedded backtick/dot/non-ASCII, spans, and the unterminated/empty rejections in both parsers; `tests/fixtures/spec/quoted-names-ok.yaml` exercises the whole path through the CLI.

Fixes #148, #137

### DIFF
--- a/crates/data-dict/src/assert_expr.rs
+++ b/crates/data-dict/src/assert_expr.rs
@@ -21,14 +21,19 @@
 //! cmp         := "=" | "!=" | "<>" | "<" | "<=" | ">" | ">="
 //! literal     := number | string | "TRUE" | "FALSE" | "NULL"
 //! funcall     := IDENT "(" (expr ("," expr)*)? ")"   // incl. NOW(), interval(n, unit)
-//! columns     := "COLUMNS" "(" ("*" | string | "[" IDENT ("," IDENT)* "]") ")"
+//! columns     := "COLUMNS" "(" ("*" | string | "[" column ("," column)* "]") ")"
 //! case        := "CASE" ("WHEN" expr "THEN" expr)+ ("ELSE" expr)? "END"
+//! column      := IDENT | QUOTED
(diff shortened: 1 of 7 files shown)


gaborcsardi:
> now go in backticks:

There are no backticks in this example. :)

gaborcsardi (approved):
Maybe we should de-duplicate the two parsers? E.g. this PR adds `parse_quoted_name()` to both, and the two implementations are exactly identical. Which makes sense, they are supposed to be identical. But then we should probably merge the two.

PR #160, opened 2026-08-02
Adopt arrow in data-dict-parquet

hadley:
Use the arrow crate instead of hand rolling so much.

- New modules: `reader.rs` (one footer parse per file, projected record-batch readers), `keys.rs` (float canonicalization + hashable key forms; the `ByteKeys` arena survives), `display.rs` (arrow's type-aware value rendering — a DATE orphan sample now reads `2024-01-02`, not `19724`).
- `column_scan.rs` (426 lines) deleted; the slow `get_row_iter` record API is gone; composite keys use `arrow-row`; decimals compare numerically via `Decimal128`/`Decimal256`, deleting `normalize_decimal`.
- The D04 page-level dictionary fast path is kept (trimmed to strings) — it's the only variant that can skip decoding data pages outright.
- Crate: 2,122 → 1,408 lines. 

## Performance

| Check | Old | New | Δ |
|-------|-----|-----|---|
| D02 uniqueness, int | 29.5–32.4 ms | 29.1–30.6 ms | parity |
| D02 uniqueness, string | 84.6–86.1 ms | 70.4–71.4 ms | **−17%** |
| D02 composite primary key | 133–135 ms | 119–121 ms | **−10%** |
| D05 foreign key, int | 59–63 ms | 62–66 ms | parity (noise) |
| D05 foreign key, string | 191–197 ms | 168–171 ms | **−13%** |
| `validate-data` peak RSS | 254 / 107 MB | 252 / 105 MB | unchanged |

String-heavy paths are 10–17% faster (arrow's byte-array decode beats the record reader)

## Cost

Binary: 6.7 → 12.6 MB unstripped (5.7 → 10.6 MB stripped)


gaborcsardi (approved):
A hard choice, but I think this is the correct one.

A few changes I would make, I think all of them should be easy.

1. Read the parquet file without reading the Arrow metadata.
   With the metadata Arrow will create different, Arrow-specific
   types, which we probably don't want. For us Arrow is just a
   tool to get to the Parquet types. This should also avoid creating
   arrow types for which we cannot create keys.

    ```
    // crates/data-dict-parquet/src/reader.rs, FileContext::open
    let meta = ArrowReaderMetadata::load(
        &file,
        ArrowReaderOptions::new().with_skip_arrow_metadata(true),
    )?;
    ```

    Maybe add a small test file written by arrow to avoid
    regression.

2. Add tests that if `uniqueness_comparability()` is true, then
   `KeyColumn::new()` actually works, because Claude thinks that
   for some types it does not, although most of this should be
   solved by not reading the metadata, i.e. the previous
   suggestion. It is a good check to keep, anyway.

3. The enum membership test always passes on an `ENUM` column,
   always, because Arrow decodes it to a binary vector (not a
   string), which we apparently ignore. This should have a test
   as well.

4. `common_type()` can return a type `KeyColumn::new()` does not
   handle: `LargeUtf8` for mixed string pairs, `LargeBinary` for
   mixed binary pairs.

5. In `check_foreign_key()`, `values.get(row)` renders more
   elements than it would be ultimately shown.

hadley:
Review feedback applied in 2884734, point by point:

1. **Skip arrow metadata** — `FileContext::open` now passes `with_skip_arrow_metadata(true)`, so types always come from the parquet schema. Regression test: `tests/arrow_written.rs` writes a file through `ArrowWriter` with `LargeUtf8` and `Dictionary` columns and asserts uniqueness still works (it errors without the option).
2. **Comparable ⇒ keyable** — `tests/comparable_keys.rs` writes every comparable parquet shape (bools, all int widths signed/unsigned, INT96, floats, float16, strings, ENUM, UUID, plain FLBA, int-/byte-/FLBA-backed decimals up to `DECIMAL(76)`, dates, times, timestamps incl. `TIMESTAMP(NANOS)`) and asserts the duplicate pair is actually found, not just that nothing errors.
3. **ENUM membership** — confirmed real bug: ENUM decodes as binary and the scan silently skipped it, so D04 could never fire. The membership check now handles binary-decoded string-likes (UTF-8 compared; non-UTF-8 rendered as hex). Tests: a true-ENUM column with a violation (`true_parquet_enum_column_is_checked`) and the dictionary-fast-path interplay.
4. **`common_type` vs `KeyColumn`** — `KeyColumn` gained `LargeUtf8`/`LargeBinary` variants, and the family table now tracks large offsets: mixed string pairs → `Utf8`/`LargeUtf8`, mixed binary pairs → `Binary`/`LargeBinary`, and string↔binary pairs (an ENUM child referencing a string parent) compare by UTF-8 bytes instead of falling into no-common-form — covered by `foreign_key_between_enum_and_string_ok`.
5. **Orphan rendering** — `ForeignKeyStats::orphan` takes the value lazily (`FnOnce`), so values render only while the ≤5-value sample can still grow, instead of for every orphan row.

PR #161, opened 2026-08-03
Add `data-dict draft`: generate a skeleton dictionary from parquet files

### LINKED ISSUE #147: `data-dict draft`: generate a skeleton data-dict.yaml from parquet files

hadley:
Closes #147.

## Changes

- **`data-dict draft <PATHS>... [-o <FILE>]`** — profiles each Parquet file and writes a starting dictionary: one table per file named after its stem, `source.parquet` relative to the output, inferred types, observed ranges and examples, and a `todo` for everything the data can't settle. Appends new tables to an existing dictionary, preserving its bytes.
- **A `todo` key at every level** — dataset, table, column, struct field, relationship. `validate-spec` reports each as an `S31` warning, so a dictionary announces its own unfinished work instead of looking finished.
- **`todo` carried through export and render** — exported as prose (Markdown rendered to HTML) at every level, and shown on the page as an alert flag whose note opens on hover; a relationship's note rides in its join tooltip.
- **A numeric column may give both `range` and `examples`** — the extremes and a typical value say different things. Each type keeps its required key and may carry the other; `string` stays examples-only. This is what lets a bare `number` draft with both, since the data can't say whether it holds ordinals or quantities (which take a range) or bare numbers (which take examples). Backwards compatible: only new combinations become legal.
- **`skill-write` → `skill-create`** — renamed, with the skill text reworked around drafting first and then resolving the todos it leaves.

## What a draft looks like

```yaml
$version: 0.1.0

todo: |
  - Write a `description` of the dataset.

tables:
  - name: readings
    source:
      parquet: readings.parquet
    todo: |
      - Record a `primary_key`, if present.
    columns:
      - name: temperature
        type: number
        range: [1.5, 9.5]
        examples: [1.5, 3, 7.25, 9.5]
        todo: |
          - Add a `description`.
          - Pick a more specific numeric type, and keep whichever of `range` and
            `examples` is appropriate.
          - Specify `constraints`. Some options suggested by the data:
            - constraints: [required] # all values present
      - name: station
        type: string
        examples: [eskdalemuir, kew, lerwick, leuchars]
        todo: |
          - Add a `description`.
          - Specify `constraints`. Some options suggested by the data:
            - constraints: [required, unique] # all values distinct & present
            - constraints: [primary_key] # all values distinct & present
            - constraints: [required] # all values present
```

Every draft validates with no errors — the only findings are the `S31` warnings its own todos raise.


hadley:
@gaborcsardi it would be most useful if you could take a look at how the TODOs are formatted (e.g. in the snapshot tests). I don't how they've turned out but I'm stuck and can't think of a better approach.

gaborcsardi (approved):

gaborcsardi:
~~The non-comment TODOs (e.g. https://github.com/tidyverse/data-dict/pull/161/changes#diff-90bad18664d76d3bdf08ebfc71144eecdc270045e57d5cebbe0956c9b8312e48R22) are also coming from `data-dict draft`?~~ Seems like it.

I guess you want to produce a valid YAML file? This is probably better than the alternative where the file does not parse until all TODOs are addressed. Or the current file does parse but it does not validate? That seems like a good solution to me. I.e. to have a valid YAML file that does not validate or at least gives warnings for the remaining TODOs. 

Otherwise I think they look good.

hadley:
Hmmmm, maybe we add a new `todo:` field that can appear anywhere and but causes a validation failure?

...

Yeah I like this approach

hadley:
I'm really liking the way this looks. I'll take another look with fresh eyes tomorrow morning.

gaborcsardi:
> Hmmmm, maybe we add a new `todo:` field that can appear anywhere and but causes a validation failure?

Great idea!

PR #162, opened 2026-08-03
Add `data-dict describe`

### LINKED ISSUE #146: `data-dict describe`: summarise the columns of a parquet file

hadley:
Closes #146.

`data-dict describe <PATH> [COLUMN] [--json]` profiles a parquet file in one pass (the #145 engine) and prints a per-column summary:

```
penguins.parquet: 9 rows × 4 columns

species — string
  Adelie     4
  Gentoo     3
  Chinstrap  2
  ------------
  distinct:  3
  missing:   0

bill_length_mm — number
  [36.7, 37.8]  ▇▇▇▇▇▇        1
  (37.8, 39.0]                0
  (39.0, 40.1]  ▇▇▇▇▇▇▇▇▇▇▇▇  2
  ...
  (58.5, 59.6]  ▇▇▇▇▇▇        1
  -----------------------------
  missing:                    2 (22.2%)
```

Each column is two sections on one grid — the data, then a rule, then the bar-less, colon-suffixed stats (`distinct:`, `missing:` with its percentage), count-aligned with the rows above.

- **Numeric / date / datetime / time** columns get a histogram.  Bars include half characters for extra resolution, with a `▏` sliver so a count below half a cell is never invisible.
- **String** columns list the 20 most common values with bare counts, then fold the hidden tail into one note with spread examples of values not shown above: `(~1180 other values, e.g. visit-0002, visit-0324, …)`.  Long labels get truncated.
- **Boolean** columns show true/false counts. , and embedded control characters render as escapes instead of breaking the row grid.
- Approximation is always marked: `distinct: ~1200` when the count is a sketch estimate, and a saturated tracker's per-value counts show as `~2` in the text and carry their `error` bound in the JSON.

## Also in this PR

- **`types parquet` is dropped**: `describe` supersedes it — the dict type is in its header, the parquet types in its `--json`. The library mapping underneath stays (describe and validate-meta's M01 use it), including a fix making it total: group and repeated fields map to `"nested"` instead of panicking in `get_physical_type`. #161 carries the same nested fix plus other overlapping CLI plumbing (and its write skill still mentions `types parquet`), so whichever PR lands second needs a small rebase.
- Integer bin widths are an engine change (`data-dict-parquet`'s profiler), not just rendering, so `draft` and any future profile consumer see the same exact edges.


hadley:
@gaborcsardi could you take a quick look? Most important is to confirm that we're not losing anything be replacing `types parquet` with `describe`, and that you think the current output looks decent.

gaborcsardi:
Well, technically we still have the parquet types in the `--json` output, but I think 1) we could have them in the summarized output as well, so that you can see the type mapping immediately, and 2) I think in the JSON output we could separate the logical type and the primitive type. The current output is:
```
❯ ./target/debug/data-dict describe ../otters/otters.parquet comments --json
{
  "path": "../otters/otters.parquet",
  "rows": 4478,
  "columns": [
    {
      "name": "comments",
      "type": "string",
      "parquet_type": "BYTE_ARRAY / String",
...
```

I don't see any reason to put them into a single `parquet_type` element.

We could also rename `type` and `parquet_type`. Instead of `type` we could use something the makes it clear that this is the data-dict type. And thinking of generalizations to databases, we could have `data_type` instead of `parquet_type` and then for the primitive type we could have another entry?

gaborcsardi:
I think the output looks very good, btw.

gaborcsardi (approved):

Issue #159, opened 2026-08-01
Enhance type system for aggregates


hadley:
Every expression node currently carries one thing, a type from `{number, string, boolean, date, datetime, interval}`. With aggregates the checker's judgement becomes a pair — **(type, shape)** — where shape is:

| Shape | Produced by | Cardinality |
|---|---|---|
| `const` | literals, `NOW()` | one value, period |
| `row` | a column reference | one value per row |
| `agg` | an aggregate call | one value per table |

The rules are small:

- An aggregate call takes a `row` (or `const`) argument and returns `agg`. Aggregates can't nest: `AVG(MIN(x))` is a shape error.
- Every existing operator and function is shape-polymorphic and takes the join of its operands' shapes, with `const` as the identity: `row ⊔ agg = row`.

`x <= MAX(x)` mixes grains and is invalid SQL, making you write `max(x) OVER ()`. But we choose to allow it:  `GROUP BY` for it to be ambiguous with, and it's a natural way to write "no value is more than twice the smallest", and it costs one extra pass.

`row` constraints are new. This would allow you to write `(max(x) - min(x)) < 10` at the cost of no longer being able to point to a specific row as the source of the failure when reporting.

## MIN/MAX force parametric signatures

Every function in the language today is monomorphic — `LENGTH: string → number`. `MIN`/`MAX` are the first `T → T`, and `T` must be constrained to the ordered types. That constraint already exists informally in the comparison operators' prose ("Ordered types: numbers, strings, dates, datetimes") but isn't a first-class thing the checker can name.

So: a minimal type-class notion, three classes, and type variables in signatures.

| Class | Members |
|---|---|
| `Ordered` | `number`, `string`, `date`, `datetime` |
| `Numeric` | `number` |
| `Any` | every type |

This is the only genuine *type* extension in the whole feature. Everything else is shape.

## Proposed initial function set

Eight functions. The bar for inclusion: it answers a question a data dictionary actually wants to state, and it has one obvious definition.

| Function | Signature | Notes |
|---|---|---|
| `ANY(b)` | `boolean → boolean` | True if any row is true. Null rows ignored. |
| `ALL(b)` | `boolean → boolean` | True if every non-null row is true. |
| `COUNT(x)` | `Any → number` | Number of non-null values. |
| `ROW_COUNT()` | `→ number` | Number of rows, nulls included. Nullary, so no `COUNT(*)` form is needed. |
[...]

PR #174, opened 2026-08-06
Aggregate expressions spec

gaborcsardi:
First step of #159.


site/expressions.md
```diff
@@ -18,11 +18,13 @@ tables:
...
 
 ## Evaluation
 
-**One row of one table.** An expression sees the columns of a single row at a time, and evaluating it against every row in turn is the whole of its meaning. There are no aggregates and no subqueries — cross-table rules belong in [`relationships`](spec.md#relationships), and the per-row restriction keeps expressions cheap to check.
+**One table, at two grains.** An expression sees the columns of one table. A bare column reference is read one row at a time, and evaluating the expression against every row in turn is the whole of its meaning; an [aggregate](#aggregate-functions) instead folds a column over every row at once, giving one value for the table. Which grain each part of an expression has is settled before any data is read — see [shapes](#shapes).
+
+There are no subqueries: cross-table rules belong in [`relationships`](spec.md#relationships).
```
hadley: Drop this?

site/expressions.md
```diff
@@ -53,16 +55,51 @@ Every expression has a type. These are the language's own types, close to but no
...
 
 The `number` measures (`number(id)`, `number(ordinal)`, `number(quantity)`) and a `datetime`'s `time_zone` do not affect type checking: all three measures are just `number`, and a `datetime` is a `datetime` whatever zone it declares.
 
+### Type classes {#type-classes}
+
+Some signatures are written over a *class* of types rather than one type:
+
+| Class     | Members                                                          |
+|-----------|------------------------------------------------------------------|
+| `Ordered` | `number`, `string`, `date`, `datetime` — the types `<` compares. |
```
hadley, line 72: Drop string from this?

site/expressions.md
```diff
@@ -303,6 +340,83 @@ All five are fixed-length. Calendar units (`months`, `years`) are deliberately e
...
+```yaml
+- assert: value <= 2 * MIN(value)
+  description: No reading is more than twice the smallest.
+```
+
+Note the mixed grain: `value` is one value per row and `MIN(value)` one per table, which [is allowed](#shapes).
+
+#### `SUM(x)`
+
+`Numeric → number`. The total of the non-null values. A sum of integers is an integer.
```
hadley, line 360: Should we make this a float?

site/expressions.md
```diff
@@ -303,6 +340,83 @@ All five are fixed-length. Calendar units (`months`, `years`) are deliberately e
...
+
+#### `AVG(x)`
+
+`Numeric → number`. The mean of the non-null values. Always a float, even over integers.
```
hadley, line 364: Technically you can average dates and date-times too, but you have to re-write `sum(x) / count()` to `min(x) + sum(x - min(x)) / count()` in order for the algebra to make sense (since `x - min(x)` becomes a duration, which you can sum.

site/expressions.md
```diff
@@ -303,6 +340,83 @@ All five are fixed-length. Calendar units (`months`, `years`) are deliberately e
...
+  description: At least 90% of customers have an email address.
+```
+
+#### `ROW_COUNT()`
+
+`→ number`. How many rows the table has, nulls included. It takes no arguments, which is why there is no `COUNT(*)` form.
+
+#### `COUNT_DISTINCT(x)`
+
+`Ordered T → number`. How many distinct non-null values `x` takes.
```
hadley, line 389: Should we an argument or different name for consider for a memory bounded approximate?

site/expressions.md
```diff
@@ -303,6 +340,83 @@ All five are fixed-length. Calendar units (`months`, `years`) are deliberately e
...
+- assert: COUNT_DISTINCT(region) <= 16
+```
+
+A function rather than a `DISTINCT` modifier.
+
+#### `ANY(b)`, `ALL(b)`
+
+`boolean → boolean`. Whether any, or every, non-null row is true.
+
+A bare `b` already asserts that `b` holds for every row, so `ALL(b)` is rarely worth writing as a whole assertion. Both earn their place inside a larger rule, where the aggregate's single value is compared or combined with something else.
```
hadley, line 401: Worth noting that these are not the same as SQL, which use `ANY` and `ALL` only with subqueries.

site/expressions.md
```diff
@@ -303,6 +340,83 @@ All five are fixed-length. Calendar units (`months`, `years`) are deliberately e
...
+
+|                            | No rows | Every value null   |
+|----------------------------|---------|--------------------|
+| `MIN`, `MAX`, `SUM`, `AVG` | null    | null               |
+| `ANY`, `ALL`               | null    | null               |
+| `COUNT`, `COUNT_DISTINCT`  | 0       | 0                  |
+| `ROW_COUNT`                | 0       | the number of rows |
+
+: {tbl-colwidths="[40,30,30]"}
+
```
hadley, line 412: Worth noting that at this behaviour is different to R? And stating what polars/pandas does?

site/expressions.md
```diff
@@ -303,6 +340,83 @@ All five are fixed-length. Calendar units (`months`, `years`) are deliberately e
...
+#### Empty and all-null input {#empty-input}
+
+Aggregates skip nulls, so a column holding nothing but nulls looks much like a table with no rows at all. What comes back is not uniform:
+
+|                            | No rows | Every value null   |
+|----------------------------|---------|--------------------|
+| `MIN`, `MAX`, `SUM`, `AVG` | null    | null               |
+| `ANY`, `ALL`               | null    | null               |
+| `COUNT`, `COUNT_DISTINCT`  | 0       | 0                  |
+| `ROW_COUNT`                | 0       | the number of rows |
```
hadley: Since the number of rows for a no rows input is 0, I think you could drop this column and just say they always return NULL.

site/expressions.md
```diff
@@ -340,17 +454,17 @@ The lambda form (`COLUMNS(c -> ...)`) and the star modifiers (`EXCLUDE`, `REPLAC
 
 ## Type checking
 
-Expressions are checked when the dictionary is validated, against the columns of the enclosing table alone — before any data is read. A malformed expression is S19, an unknown column S20, an ill-typed expression S21, and an empty column selection S22; see [validation](validation.md) for the full list.
+Expressions are checked when the dictionary is validated, against the columns of the enclosing table alone — before any data is read. A malformed expression is S19, an unknown column S20, an ill-typed expression S21, an empty column selection S22, and a nested aggregate S30; see [validation](validation.md) for the full list.
```
hadley: Not new here, but I think we should drop the validation rule names from this doc (I think I accidentally added them)

site/expressions.md
gaborcsardi: Hmmm. `string` is Ordered, no?

site/expressions.md
gaborcsardi: IDK. We'll only need to decide this when we start to run the checks, but indeed we'll need to do something with it.

site/expressions.md
gaborcsardi: IDK, I am not really convinced that current data-dict works well for data sets that do not fit into memory, and especially when even a single column does not fit comfortably, so it does not feel very important to me.

site/expressions.md
gaborcsardi: We have this in the spec, so I am afraid this does not work currently:
> Subtracting one date from another is not supported, nor is multiplying an interval.

site/expressions.md
gaborcsardi: Yeah. Should we try to switch to the R/python behavior instead?

gaborcsardi: Actually, probably not. In particular it'd cause issues with strings and would not buy us much. I'll add a note that it is different.

site/expressions.md
gaborcsardi: Actually, we need string here, otherwise things like COUNT_DISTINCT() won't work on strings.

site/expressions.md
gaborcsardi: we could also make it a parameter of COUNT_DISTINCT()?

gaborcsardi:
OK, I updated the PR. Two outstanding issues that we can handle later:
- memory bounded `COUNT_DISTINCT()`
- what to do with integer SUM()

gaborcsardi:
Are you ok with merging this? Then I'll merge the check part as well....

hadley (approved):
Sorry, yes go for it!

PR #175, opened 2026-08-06
Validate aggregate expressions

### LINKED ISSUE #159: Enhance type system for aggregates

gaborcsardi:
Closes #159.


crates/data-dict/src/assert_expr.rs
```diff
@@ -927,13 +929,109 @@ fn kind_to_ty(kind: ColumnKind) -> Ty {
...
+            ret,
+            aggregate: true,
+        }
+    }
+}
+
+/// The signature of the named function, or `None` if there is no such function.
+/// `name` must already be lowercased. `NOW` and `interval` are absent: the
+/// parser gives them their own AST nodes.
+fn signature(name: &str) -> Option<Sig> {
```
hadley, line 1012: I was surprised that this didn't declare the shape too.

hadley (approved):

crates/data-dict/src/assert_expr.rs
gaborcsardi: Yeah, it is essentially the `aggregate` flag in `Sig`. Maybe we could have an enum instead?

Issue #158, opened 2026-08-01
Proposal: executing expressions — evaluate against Parquet, translate to R/Python/SQL


hadley:
Today an `assert` expression is parsed and type-checked when the dictionary is validated (S19–S23), but then nothing ever runs it. There are two places we should need this:

1. **`validate-data` should check assertions.** The dictionary says `end_date >= start_date`; the data level should report the rows where that is false, exactly as it reports nulls in a `required` column (D01) or values outside an enum (D04). (#126)
2. **Users should be able to run the same checks in their own stack.** A data dictionary outlives any one tool. A team living in the tidyverse, data.table, pandas, polars, or a database wants the dictionary's rules as native code — to paste into a pipeline, wire into a test suite, or install as a `CHECK` constraint — without hand-translating each expression.

Both are functions of the same thing. The dialects proposal established the principle: **one language, one typed AST, and everything downstream is dialect-blind**. Input dialects and output targets are symmetric around that AST:

```text
source text (SQL / R / Python spellings)
  │ parse
  ▼
surface AST
  │ check + lower
  ▼
typed IR
  ├─▶ interpreter (Parquet, validate-data)
  ├─▶ SQL emitters (ANSI, DuckDB, PostgreSQL)
  ├─▶ R emitters (base, tidyverse, data.table)
  └─▶ Python emitters (polars, pandas)
```

The interpreter is the **reference implementation** of [expressions.md](https://data-dict.tidyverse.org/expressions.html): it implements the spec exactly. Emitted code is required to agree with it wherever the target language can, with every divergence classified and surfaced (see [fidelity](#the-fidelity-contract)).

## The typed IR

Today `check()` type-checks the surface AST and discards the types. The missing piece is a lowering step that runs after a clean check and produces a **typed IR** — the single contract every backend consumes:

* **Every node carries its type** (`number`, `string`, `boolean`, `date`, `datetime`, `interval`), so no backend re-derives typing rules.
* **Column references are resolved** to the column's declared type (and, for the interpreter, its position in the file's schema).
* **Function names are resolved to canonical operations.** The surface AST's stringly `Call { name }` (and every dialect alias) becomes one variant per function; emitters match on an enum, never on spelling.
[...]

PR #181, opened 2026-08-07
Expression execution spec

gaborcsardi:
First step for #158.

### DIFF
--- a/site/expression-execution.md
+++ b/site/expression-execution.md
@@ -146,8 +146,8 @@ A translated assertion is a predicate; what a caller usually wants is the rows t
 | Target | Violating rows | Why |
 |--------|----------------|-----|
 | `SQL` | `SELECT * FROM t WHERE (expr) IS FALSE` | `IS FALSE` is false-only by definition; null and true both escape. |
-| `R(base)` | `t[which(!(expr)), ]` | `which()` keeps only `TRUE`; `!NA` is `NA` and drops out. |
-| `R(tidyverse)` | `filter(t, !(expr))` | `filter()` keeps only `TRUE`. |
+| `R(base)` | `subset(t, !(expr))` | `subset()` keeps only `TRUE`; `!NA` is `NA` and drops out. |
+| `R(tidyverse)` | `filter(t, !(expr))` | `filter()` keeps only `TRUE`; `!NA` is `NA` and drops out. |
 | `R(data.table)` | `t[!(expr)]` | `NA` in `i` selects nothing. |
 | `Python(pandas)` | `df.query("not (expr)")` | `query()` keeps only `True`; on nullable dtypes a null is not `True`. |
 | `Python(polars)` | `df.filter(~(expr))` | `~` maps false to true and null to null; `filter` drops null. |
@@ -160,6 +160,8 @@ These are documentation, not output — how to embed a predicate is the caller's
 
 `translate` is primarily for machine consumption, so it writes JSON to standard output: one record per expression, carrying the source text, where it came from, the expression's type, the columns it uses, and one entry per target.
 
+A translation's `fidelity` is the weakest [class](#fidelity) among the constructs it used, written `"exact"`, `"guarded"`, `"divergent"` or `"unsupported"`. The field is present only when it isn't `"exact"`, and `notes` only when `fidelity` is. An entry carrying neither is exact and has nothing to warn about.
+
 ```json
 {
   "expr": "LENGTH(postcode) <= 10",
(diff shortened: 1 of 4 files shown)


gaborcsardi:
`SQL(ANSI)` seems to bring up more issues than the other languages, we are essentially creating our own SQL dialect, which smells a bit. Maybe we could postpone it in the first round?

site/expression-execution.md
```diff
@@ -0,0 +1,304 @@
...
+
+### One expression at a time
+
+The unit of translation is one expression. By default every assertion in the dictionary is translated; `--table` narrows that to one table, and `--expr` translates an ad-hoc expression instead.
+
+An `--expr` expression is parsed, resolved and type-checked exactly like an assertion, with one relaxation: it need not be boolean. `a + b` translates, and its type is reported. Its column names resolve against one table — the only table if the dictionary has one, and otherwise the table named by `--table`.
+
+## `SQL(ANSI)` {#ansi}
+
+`SQL(ANSI)` is the portable SQL target: the one to use when the destination engine isn't known, or when the same text has to run on more than one.
```
hadley, line 192: I think this is just claude complicating things. There is not 100% spec conforming ANSI SQL engine but I think we can assume that if it works in Postgres and works in duckdb it's valid ANSI SQL (that's very likely to be true for the subset of SQL we're dealing with). 

It would also be fine to start with either duckdb or SQLite as the initial SQL dialect

site/expression-execution.md
```diff
@@ -0,0 +1,304 @@
...
+DuckDB is the interesting case, since the language's `COLUMNS` is modelled on DuckDB's. DuckDB's is a syntactic macro rather than a value: it rewrites the enclosing expression once per column, and where the results are combined with `AND` depends on the clause it appears in. Since `translate` returns a bare predicate whose eventual context it cannot see, keeping the symbolic form would let `WHERE (COLUMNS(*) IS NOT NULL) IS FALSE` mean "every column is null" instead of "some column is null". DuckDB therefore expands, with a note.
+
+### Embedding a predicate {#embedding}
+
+A translated assertion is a predicate; what a caller usually wants is the rows that break it. Since an assertion passes on true **or** null, the violating rows are exactly those where the predicate is `false` — which is not the same as "not true". Each target has a native way to say it:
+
+| Target | Violating rows | Why |
+|--------|----------------|-----|
+| `SQL` | `SELECT * FROM t WHERE (expr) IS FALSE` | `IS FALSE` is false-only by definition; null and true both escape. |
+| `R(base)` | `t[which(!(expr)), ]` | `which()` keeps only `TRUE`; `!NA` is `NA` and drops out. |
```
hadley: `subset`?

site/expression-execution.md
```diff
@@ -0,0 +1,304 @@
...
+| `R(tidyverse)` | `filter(t, !(expr))` | `filter()` keeps only `TRUE`. |
```
hadley: Now `filter_out()`

site/expression-execution.md
```diff
@@ -0,0 +1,304 @@
...
+{
+  "expr": "LENGTH(postcode) <= 10",
+  "table": "survey",
+  "type": "boolean",
+  "columns": [{ "table": "survey", "column": "postcode" }],
+  "translations": [
+    { "target": "R(base)", "code": "nchar(postcode) <= 10", "notes": [] },
+    { "target": "R(tidyverse)", "code": "str_length(postcode) <= 10", "notes": [] },
+    { "target": "SQL(duckdb)", "code": "length(\"postcode\") <= 10", "notes": [] },
+    { "target": "Python(pandas)", "code": "postcode.str.len() <= 10",
```
hadley, line 175: Maybe write out `fidelity` when not "exact", and then only include the notes field in that case.

site/expression-execution.md
gaborcsardi: Well, I created a grammar following this heuristic, so we can check against that grammar.

hadley: Also fine to postpone. The grammar just feels a bit overwrought?

site/expression-execution.md
gaborcsardi: Hmmm, `filter_out()` keeps both `NA` and `FALSE`, which is not what we want here, I think.

hadley: From the docs we have `end_date >= start_date`, and that assertions state what must be true. So I think that implies that the data you want to analyse is `df |> filter(end_date >= start_date)` and that everything you want to report is `df |> filter_out(end_date >= start_date)`?

hadley: Maybe implies that the NA rules for aggregates are wrong?

gaborcsardi: We do not want to report the `NA` rows. I.e. a constraint on a missing value is OK by default. So that means we need `filter()`, right? `df |> filter_out(end_date >= start_date)` will keep rows where at least one of `end_date` or `start_date` is missing as violating rows. Which is not what we want. No?

gaborcsardi:
The other thing about `SQL(ANSI)` is that if that's a SQL dialect that works on DuckDB (as well), then why are we creating another dialect that only works on DuckDB?

hadley:
I was thinking of it as the "generic" dialect that all others build off, and the one you'd start with for a database that didn't have its own dialect. So it's more like the lowest common denominator than trying to be fully ANSI sql compliant.

hadley (approved):

PR #185, opened 2026-08-07
Expression translation

gaborcsardi:
Expression interpreter, with translation CLI and DuckDB support.


crates/data-dict/src/assert_expr/ir.rs
```diff
@@ -0,0 +1,1090 @@
...
+            | NodeKind::Bool(_)
+            | NodeKind::Null
+            | NodeKind::Date(_)
+            | NodeKind::Datetime(_)
+            | NodeKind::Column(_)
+            | NodeKind::Selected
+            | NodeKind::Now => Vec::new(),
+            NodeKind::Neg(x) | NodeKind::Not(x) => vec![x],
+            NodeKind::Arith { lhs, rhs, .. } | NodeKind::Compare { lhs, rhs, .. } => vec![lhs, rhs],
+            NodeKind::And(l, r) | NodeKind::Or(l, r) => vec![l, r],
```
hadley: lhs rhs for consistency?

crates/data-dict/src/assert_expr/ir.rs
```diff
@@ -0,0 +1,1090 @@
...
+    },
+    Between {
+        operand: Box<TypedExpr>,
+        lo: Box<TypedExpr>,
+        hi: Box<TypedExpr>,
+        negated: bool,
+    },
+    In {
+        operand: Box<TypedExpr>,
+        list: Vec<TypedExpr>,
```
hadley:
```suggestion
        needle: Box<TypedExpr>,
        haystack: Vec<TypedExpr>,
```

I always find those terms evocative.

crates/data-dict/src/assert_expr/ir.rs
```diff
@@ -0,0 +1,1090 @@
...
+    Exact(String),
+    /// `'abc%'`
+    Prefix(String),
+    /// `'%abc'`
+    Suffix(String),
+    /// Anything else, as an anchored regex.
+    Regex(String),
+    /// A computed pattern. A target that must compile the pattern at
+    /// translation time can't support this one.
+    Dynamic(Box<TypedExpr>),
```
hadley, line 272: This is like when you use a column value?

crates/data-dict/src/assert_expr/ir.rs
```diff
@@ -0,0 +1,1090 @@
...
+/// One variant per function in the language. Backends match on this; the
+/// spelling that produced it, and its case, are gone.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Op {
+    Length,
+    Lower,
+    Upper,
+    Trim,
+    StartsWith,
+    EndsWith,
```
hadley, line 284: Given we have Compare, Arith etc, I wonder if it would be useful to break this down into smaller pieces? Like NumericOp, StringOp, DateOp?

crates/data-dict/src/assert_expr/ir.rs
```diff
@@ -0,0 +1,1090 @@
...
+    Mod,
+    Min,
+    Max,
+    Sum,
+    Avg,
+    Count,
+    RowCount,
+    CountDistinct,
+    Any,
+    All,
```
hadley, line 298: Maybe Aggregate for these?

crates/data-dict/src/assert_expr/ir.rs
```diff
@@ -0,0 +1,1090 @@
...
+                TypedExpr {
+                    ty: Type::Bool,
+                    shape: lhs.shape.max(rhs.shape),
+                    kind: NodeKind::Compare { op: *op, lhs, rhs },
+                    span,
+                }
+            }
+            ExprKind::IsNull { operand, negated } => {
+                let operand = self.boxed(operand)?;
+                TypedExpr {
```
hadley, line 512: I guess I'm surprised that this doesn't convert `negated` to NOT here? I guess that must happen later.

crates/data-dict/src/assert_expr/ir.rs
```diff
@@ -0,0 +1,1090 @@
...
+fn lit(kind: NodeKind, ty: Type, span: (usize, usize)) -> TypedExpr {
+    TypedExpr {
+        kind,
+        ty,
+        shape: Shape::Const,
+        span,
+    }
+}
+
+/// A `CASE`'s type is its branches' common type, or [`Type::Any`] when they
```
hadley, line 775: Why not error for mixed types here?

crates/data-dict/src/assert_expr/ir.rs
```diff
@@ -0,0 +1,1090 @@
...
+            Some(_) => Type::Any,
+        });
+    }
+    result.unwrap_or(Type::Any)
+}
+
+/// Turn a string literal that met a temporal operand into a real date or
+/// datetime, so a backend constructs one natively rather than relying on the
+/// target's own string coercion. This is the comparability rule that admits
+/// `birthdate >= '2000-01-01'`, applied to the value rather than the type.
```
hadley:
```suggestion
```

crates/data-dict/src/assert_expr/ir.rs
```diff
@@ -0,0 +1,1090 @@
...
+        // Mixed grain: a row-level operand widens the whole thing back to row.
+        assert_eq!(ir("qty <= 2 * MIN(qty)").root.shape, Shape::Row);
+        assert_eq!(ir("1 > 0").root.shape, Shape::Const);
+    }
+
+    #[test]
+    fn min_returns_its_argument_type() {
+        assert_eq!(
+            rendered("d >= MIN(d)"),
+            "Ge(col(d):date, Min(col(d):date):date):bool"
```
hadley, line 1025: I wonder if it's worth making a little tree rendered so we could emit something like this in snapshot tests:

```
Ge:bool
  col(d):date
    d
  Min:date
    col(d):date
```

crates/data-dict/src/emit/duckdb.rs
```diff
@@ -0,0 +1,322 @@
...
+};
+
+pub struct DuckDb;
+
+impl Target for DuckDb {
+    fn name(&self) -> &'static str {
+        "SQL(duckdb)"
+    }
+
+    fn prec(&self, e: &TypedExpr) -> u8 {
```
hadley, line 21: It would be useful to have a comment somewhere just generally explaining how this precedence transformation works.

crates/data-dict/src/emit/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+            self.notes.insert(note);
+        }
+    }
+
+    /// The column `Selected` stands for right now.
+    pub fn selected(&self) -> Option<&ColumnRef> {
+        self.selected
+    }
+
+    /// Emit `e` as an operand of an operator with precedence `parent`,
```
hadley, line 159: ok, this is what I was looking for 😄

crates/data-dict/src/emit/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    #[test]
+    fn columns_are_quoted() {
+        assert_eq!(sql("qty > 0"), r#""qty" > 0"#);
+        // A struct field is a path, each segment quoted.
+        assert_eq!(sql("LENGTH(addr.zip) > 0"), r#"length("addr"."zip") > 0"#);
+    }
+
+    #[test]
+    fn parentheses_appear_only_where_they_change_the_reading() {
```
hadley, line 275: It would be nice to have a some examples where the output gains parentheses

hadley (approved):
I can't say that I've reviewed this in detail, but the overall shape is certainly what I expected

crates/data-dict/src/assert_expr/ir.rs
gaborcsardi: Exactly. Any time it is not a literal string.

crates/data-dict/src/assert_expr/ir.rs
gaborcsardi: It is actually simpler and faster to keep it a flat table.

crates/data-dict/src/assert_expr/ir.rs
gaborcsardi: Sounds like a good improvement for later!

crates/data-dict/src/assert_expr/ir.rs
gaborcsardi: This only happens in the backend, because it is different on every backend.

crates/data-dict/src/assert_expr/ir.rs
gaborcsardi: because sometimes you want to keep different types that are comparable by converting them later. E.g. a date and a datetime.

PR #187, opened 2026-08-07
Tidyverse translation support

gaborcsardi:


crates/data-dict/src/emit/r_tidyverse.rs
```diff
@@ -0,0 +1,501 @@
...
+    ArithOp, CmpOp, DatetimeConst, IntervalUnit, LikePattern, NodeKind, Op, Selection,
+    SelectorForm, Type, TypedExpr,
+};
+
+pub struct RTidyverse;
+
+/// R's precedence, which differs from SQL's in one place that matters: an
+/// infix `%…%` operator binds *tighter* than `*` and `/`, so `%in%` has a level
+/// of its own above them.
+mod p {
```
hadley, line 21: I presume this will eventually get hoisted up to an R base layer when we add base/data.table support.

crates/data-dict/src/emit/r_tidyverse.rs
```diff
@@ -0,0 +1,501 @@
...
+    fn name(&self) -> &'static str {
+        "R(tidyverse)"
+    }
+
+    fn prec(&self, e: &TypedExpr) -> u8 {
+        match &e.kind {
+            NodeKind::Or(..) => p::OR,
+            // A guarded `IN` is an `|`, and a guarded `MOD` is a subtraction:
+            // both sit where their emitted form sits, not where the language's
+            // operator would.
```
hadley, line 43: I don't know what this means

crates/data-dict/src/emit/r_tidyverse.rs
```diff
@@ -0,0 +1,501 @@
...
+    pub const NEG: u8 = 8;
+    pub const ATOM: u8 = 9;
+}
+
+impl Target for RTidyverse {
+    fn name(&self) -> &'static str {
+        "R(tidyverse)"
+    }
+
+    fn prec(&self, e: &TypedExpr) -> u8 {
```
hadley, line 38: I assume the overall goal here is to align the precedences so that we insert parentheses where needed.

crates/data-dict/src/emit/r_tidyverse.rs
```diff
@@ -0,0 +1,501 @@
...
+            // `matches()` is unanchored, like the language's own regex.
+            SelectorForm::Regex(pattern) => cx.push(&format!("matches({})", string(pattern))),
+            SelectorForm::List => {
+                cx.push("c(");
+                for (i, column) in selection.columns.iter().enumerate() {
+                    if i > 0 {
+                        cx.push(", ");
+                    }
+                    cx.push(&self.column(&column.path));
+                }
```
hadley: Probably worth a helper for this

crates/data-dict/src/emit/r_tidyverse.rs
```diff
@@ -0,0 +1,501 @@
...
+    "R folds an empty or all-null column to the identity (0, FALSE, TRUE, Inf) where data-dict returns null, so an aggregate assertion differs on such a column.",
+);
+
+const OVERFLOW: Fidelity = Fidelity::Divergent(
+    "R has no 64-bit integers, so arithmetic data-dict reports as an overflow (D09) yields a double here.",
+);
+
+/// Whether this arithmetic is a temporal shift, which needs the operand
+/// promoted before the duration is added.
+fn is_shift(lhs: &TypedExpr, rhs: &TypedExpr) -> bool {
```
hadley: I think `is_temporal()` would be clearer than `is_shift()`

crates/data-dict/src/emit/r_tidyverse.rs
```diff
@@ -0,0 +1,501 @@
...
+    lhs.ty == Type::Interval || rhs.ty == Type::Interval
+}
+
+/// A `Date` plus a duration stays a `Date` in R, silently dropping anything
+/// shorter than a day; the language makes it a datetime. Promoting the date
+/// first restores that, so the mapping is exact rather than approximate.
+fn write_shift(
```
hadley: Same here: `write_temporal()`

crates/data-dict/src/emit/r_tidyverse.rs
```diff
@@ -0,0 +1,501 @@
...
+        if i > 0 {
+            cx.push(", ");
+        }
+        cx.free(item)?;
+    }
+    cx.push("))");
+    Ok(())
+}
+
+fn write_like(
```
hadley, line 306: I think this could be simplified to use `stringr::str_like()` 😄

crates/data-dict/src/emit/r_tidyverse.rs
```diff
@@ -0,0 +1,501 @@
...
+            cx.free(&args[0])?;
+            cx.push(", fixed(");
+            cx.free(&args[1])?;
+            cx.push("))");
+        }
+        Op::Abs => cx.call("abs", &refs)?,
+        Op::Floor => cx.call("floor", &refs)?,
+        Op::Ceil => cx.call("ceiling", &refs)?,
+        Op::Round => {
+            cx.fidelity(ROUNDING);
```
hadley, line 381: I really like the way that fidelity is reported.

crates/data-dict/src/emit/r_tidyverse.rs
```diff
@@ -0,0 +1,501 @@
...
+            cx.call("round", &refs)?;
+        }
+        // R's `%%` takes its sign from the divisor; the language takes it from
```
hadley, line 384: IMO I think we should adopt the R/python convention not the SQL/C convention

hadley (approved):

crates/data-dict/src/emit/r_tidyverse.rs
gaborcsardi: Yeah, that's terrible lingo all over... "guarded" means that we cannot just write `%IN%` because it behaves differently for `NA`, so we need to write `is.na(x) | x %in% c(....)`. So an `IN` becomes a (vectorized) `|`. 

The `MOD` is actually a bug, possibly, let me take a closed look.

gaborcsardi: Yeah, so `MOD` is `x - y * trunc(x / y)`, so it is a minus.

crates/data-dict/src/emit/r_tidyverse.rs
gaborcsardi: YEs, exactly.

crates/data-dict/src/emit/r_tidyverse.rs
gaborcsardi: It is not just about being temporal but one needs to be a duration specifically. Maybe `is_interval_shift()` ?

crates/data-dict/src/emit/r_tidyverse.rs
gaborcsardi: I made this `write_interval_shift()`.

crates/data-dict/src/emit/r_tidyverse.rs
gaborcsardi: Seems like `str_like()` is not doing enough, e.g. it does not escape some regex characters?
```r
> stringr:::like_to_regex("a.c%")
[1] "^a.c.*$"
str_like(c("a.c", "abc"), "a.c%")
#> [1]  TRUE  TRUE
```
The second one should not match here.

crates/data-dict/src/emit/r_tidyverse.rs
gaborcsardi: OK. Let's do this in a separate PR, it also affects the spec, not just tidyverse.

gaborcsardi: Done in #191.

crates/data-dict/src/emit/r_tidyverse.rs
hadley: Or `is_duration()`?

crates/data-dict/src/emit/r_tidyverse.rs
hadley: https://github.com/tidyverse/stringr/issues/617

crates/data-dict/src/emit/r_tidyverse.rs
gaborcsardi: Well, we are checking if one side is a duration, i.e. the _operation_ is a shift?

hadley: Oh, gotcha.

PR #188, opened 2026-08-07
Diagram layout, a way back, and a favicon for the rendered page

paul-dicristina:
Six independent changes to the page `data-dict render` writes. Each commit stands alone, so any of them can be dropped.

### The diagram

- **A quarter taller.** Both terms of the canvas height are scaled, so the board gains the room on short viewports (where the `vh` term wins) and tall ones (where the pixel cap does).
- **The arrangement is remembered.** Dragging a table into place is deliberate, so it now outlives the page: positions and the board's pan go to `localStorage`, keyed by the dictionary rather than the URL, since the same page is served from a port under `--live` and opened from disk once written out. Only a drag rewrites the arrangement, so panning around a layout an eye rebuilt can't quietly replace the one you arranged by hand. `tidy` discards it.
- **A table opens from its heading.** A chevron after the name, and the two together link to that table's page — the diagram's names were the one place a table couldn't be reached from.

### The page

- **The dictionary's name is the way back.** Inside a table the title is itself the link home, led by a chevron, replacing the small `home` link beside it. The chevron hangs in the left margin so the name keeps the position it has on every other page: navigating home doesn't slide the title sideways.
- **Description cards sit on a warm surface**, so prose stops reading as the same white sheet as the column list beside it.
- **A favicon**, inlined as a data URI so the page stays one self-contained file.

### Two things worth a reviewer's attention

**The diagram heading is both a drag handle and a link.** That combination is awkward: dragging captures the pointer to the heading, and a captured pointer delivers its click *there* rather than to the link inside it, so the link is never activated by pointer. A press that ends where it began therefore follows the link itself, and one that moved is a drag and opens nothing. The `href` stays real, so keyboard activation and a modified click for a new tab both go through the browser rather than through this code.

**None of this is covered by a test.** There is no browser harness in the repo, and the `render` tests in `tests/cli.rs` assert on emitted HTML, which can't reach drag, storage, or pointer behaviour. I verified all of it by driving headless Chrome over CDP — click-opens-table, drag-does-not, click-after-drag, `Enter` on the focused link, the eye still toggling, dragging by the column count, pan and arrangement surviving a reload, `tidy` forgetting them, and the title holding position at both breakpoints — but those scripts aren't in the repo. Happy to add a harness, or a cheaper smoke test, if you'd like either before merging.


hadley:
Thanks!

PR #194, opened 2026-08-10
Expressions: divide by zero is now Inf

### LINKED ISSUE #193: Expressions: should division by zero be +-Inf?

gaborcsardi:
- Update spec to include infinity and NaN, including two new literals.
- New IS_FINITE(), IS_INFINITE(), IS_NAN().
- Update execution spec.
- MOD(x, 0) is NaN.
- Update ordering: -INF < finite < INF < NaN. All NaN one value.
- D10 not needed any more.
- NaN rejected as a bound.

For now this is the spec only, no code changes.

Closes #193 (well, once actually implemented).


site/expression-execution.md
```diff
@@ -27,18 +27,26 @@ A **mixed-grain** assertion such as `value <= 2 * MIN(value)` is a `row` asserti
...
 
 This does mean evaluation is not total — some data can stop an assertion from reaching a verdict. That is a deliberate trade. A rule that cannot be computed has not been checked, and saying so is more useful than a pass nobody earned.
 
+#### Non-finite values {#non-finite}
+
+Division by zero is an infinity or a NaN, [as the language specifies](expressions.md#non-finite): `7 / 0` is `INF`, `0 / 0` is a NaN, `MOD(x, 0)` is a NaN. Nothing is reported, and evaluation carries on.
+
+This is safe because comparing anything against a NaN gives `false`, so the row is reported as a violation. Giving null instead would be unsafe, because [null passes](#what-counts-as-a-violation): `total / qty > 1` would then go unchecked on exactly the rows whose `qty` is most suspect.
```
hadley: Shouldn't `NaN` propagate like `NULL`? i.e. `NaN < 5` is `NULL`?

hadley (commented):
I think my main question is if we're over specifying the behaviour here. I worry a bit that by fully specifying the behaviour, which requires a lot of text, we inadvertently draw the LLM to overfixate on this behaviour, which we don't generally believe to be that important.

site/expression-execution.md
gaborcsardi: That's not good for us (at least here), because `NULL` passes constraints, so it is not a violation. I assume we want a violation for an `NaN < 5` constraint?

gaborcsardi:
> I think my main question is if we're over specifying the behaviour here. I worry a bit that by fully specifying the behaviour, which requires a lot of text, we inadvertently draw the LLM to overfixate on this behaviour, which we don't generally believe to be that important.

IDK. We can certainly make it less verbose. But we need to define it unambiguously _somewhere_ to implement it, no?

hadley:
Yeah, that's true. So maybe what I'm actually asking is that we pull all of this out into a single file that defines our FP behaviour.

gaborcsardi:
OK, I separated it out. Full IEEE 754 support certainly seems like the correct thing to do, in retrospect.

@hadley what do you think?

site/floating-point.md
```diff
@@ -0,0 +1,70 @@
...
+
+Two spellings that are otherwise interchangeable stop being so here. `NOT (x < y)` and `x >= y` differ when either side is a NaN: the first is `true`, the second `false`. Where a column can hold one, prefer the positively-stated form, or conjoin `IS_FINITE(...)`.
+
+### Equality is not identity {#identity}
+
+`=` is IEEE, so no NaN equals any NaN. But several places have to decide whether two values are *the same value* rather than whether they compare equal, and there the language uses one **identity order** instead: values run `-INF` < every finite number < `INF` < NaN, all NaNs count as one value, and `-0.0` and `+0.0` count as one value.
+
+The identity order governs `MIN`, `MAX` and `COUNT_DISTINCT` in an expression, and the `unique`/`primary_key` ([D02](validation.md#data-validation-checks)) and `foreign_key` ([D05](validation.md#data-validation-checks)) checks at [the data level](validation.md#comparable-types). So `COUNT_DISTINCT` of a column of nothing but NaNs is 1; `MIN` of a column containing a NaN is its smallest ordinary value, and `MAX` is the NaN.
```
hadley, line 50: MAX being NaN surprises me. I think they shouldn't be orderedable so are dropped from MIN/MAX.

site/floating-point.md
```diff
@@ -0,0 +1,70 @@
...
+
+`SUM` and `AVG` are arithmetic rather than ordering, so they follow IEEE and propagate: one NaN anywhere in the column makes both a NaN, and an `INF` and a `-INF` in the same column sum to a NaN. `COUNT` counts a NaN, because a NaN is not null.
+
+## Evaluation {#evaluation}
+
+Division by zero is an infinity or a NaN: `7 / 0` is `INF`, `0 / 0` is a NaN, `MOD(x, 0)` is a NaN. Nothing is reported, and evaluation carries on.
+
+This is safe because comparing anything against a NaN gives `false`, so the row is reported as a violation. Giving null instead would be unsafe, because [null passes](expression-execution.md#what-counts-as-a-violation): `total / qty > 1` would then go unchecked on exactly the rows whose `qty` is most suspect.
```
hadley, line 58: Not `NULL`?

site/floating-point.md
```diff
@@ -0,0 +1,70 @@
...
+
+Integer overflow is the one thing that still has no answer, and it is [reported as D09](expression-execution.md#no-result). Floats are unaffected: they overflow to `INF`, which is a value, so the expression still reaches a verdict.
+
+A NaN or an infinity in the data is treated the same way as one the arithmetic produced. `validate-data` reads a float column as it finds it, so such a value is a value, not a missing one and not a broken column: `IS NULL` is `false` for it, `required` is satisfied by it ([D01](validation.md#data-validation-checks)), `COUNT` counts it, an aggregate folds it in, and any comparison against a NaN is `false`.
+
```
hadley: Delete this?

site/floating-point.md
```diff
@@ -0,0 +1,70 @@
...
+The profile treats the same values differently, on purpose: it counts them [separately](export.md#profile) instead of binning them. A profile shows where values sit on the number line, and an infinity has no place there — it would stretch every bin. An assertion just asks whether a rule holds, and there an infinity is an ordinary value.
+
+## In a dictionary {#in-a-dictionary}
+
+A `range` bound may be `-.inf` or `.inf` to [leave that end open](spec.md#representative-values). That is a statement about the bound, not about the data: it says the true extent is unknown or moving, rather than that the column was observed to hold an infinity. `.nan` is not a bound at all and is rejected (S12).
+
+R, Python and SQL disagree about all of this more than about anything else in the language — over what `x / 0` gives, over whether a NaN equals itself, and over whether a NaN counts as missing. [Executing expressions](expression-execution.md#standing-divergences) has the table.
```
hadley: Put the table here?

site/floating-point.md
gaborcsardi: IDK what is more surprising, if it is dropped or not. It is the same in R:
```r
> max(NaN)
[1] NaN
> max(c(1,NaN))
[1] NaN
```
and also in duckdb. Not in numpy, though, but then that breaks the common ordering rule.

site/floating-point.md
gaborcsardi: `NaN` value is a violation, `NULL` is not. If that's what you are asking.

hadley: I meant like this:

```
> 10 > NaN
[1] NA
```

site/floating-point.md
hadley: Oh ok, that's fine then. I guess these values are so rare it's not really that important.

hadley: Oh but that's because R treats it like an NA:

```R
> min(c(1, NaN))
[1] NaN
```

site/floating-point.md
gaborcsardi: If we make `10 > NaN` evaluate to `NULL` , then we are not going to report that row as a violation. But we do want to report these, no?

hadley: That's related to my other comment where I think `NULL` should be a validation error. i.e. if you say `x < 10` and you don't know what `x` is, then I don't think the constraint is upheld.

hadley (approved):
Makes sense!

Issue #134, opened 2026-07-23
Define metrics/filters/examples


hadley:
i.e. include some useful operations 

```yaml
$version: 0.2.0
name: sales
tables:
  - name: orders
    description: One row per completed order since 2020.
    source:
      sql: sales.orders          # already planned for the spec
    columns:
      - name: order_total
        type: number(quantity)
        units: USD
        range: [0, .inf]
      - name: region
        type: enum
        values: [AMER, EMEA, APAC]
      - name: status_cd
        type: enum
        values: {90: realized, 10: pending, 20: cancelled}
    metrics:                               # NEW: governed aggregations
      - name: net_revenue
        label: Net revenue
        description: Realized revenue net of discounts. Only status 90 counts.
        expr: SUM(order_total * adj_fctr) FILTER (WHERE status_cd = 90)
        origin:                            # optional pointer to the system of record
          snowflake_semantic_view: SALES_DB.PUBLIC.SALES_METRICS
          metric: NET_REVENUE

filters:                                   # NEW: named segments
  - name: enterprise
    description: Mid-Market-3 through Enterprise-3 accounts.
    expr: tile_size IN ('Mid-Market-3', 'Enterprise-1', 'Enterprise-2', 'Enterprise-3')

queries:                                   # NEW: verified, parameterized queries
  - name: open_opps_in_region
    description: Open opportunities in the given regions.
    parameters:
      - name: regions
        type: enum
        values: [AMER, EMEA, APAC]
    sql: |
      SELECT * FROM opportunities
      WHERE status = 'open' AND region IN ({regions})
    verified_by: jane
    verified_at: 2026-07-01
```

hadley:
Maybe we could use a generic computation field?

```yaml
$version: 0.2.0
name: sales
tables:
  - name: orders
    description: One row per completed order since 2020.
    source:
      sql: sales.orders          # already planned for the spec
    columns:
      - name: order_total
        type: number(quantity)
        units: USD
        range: [0, .inf]
      - name: region
        type: enum
        values: [AMER, EMEA, APAC]
      - name: status_cd
        type: enum
        values: {90: realized, 10: pending, 20: cancelled}
    computations:                               # NEW: governed aggregations
      - name: net_revenue
        label: Realized revenue net of discounts
        description: Only status 90 counts.
        expr: SUM(order_total * adj_fctr) FILTER (WHERE status_cd = 90)
      - name: enterprise
        description: > 
[...]

PR #202, opened 2026-08-10
Spec for a new `definitions` field

hadley:
Part of #134


site/spec.md
```diff
@@ -353,6 +354,36 @@ constraints:
 
 [Expressions](expressions.md) documents the language in full: every operator and function with its input and output types, precedence, the `COLUMNS(...)` forms, the type rules a validator enforces, and the grammar.
 
+### Definitions
+
+A table's `definitions` property is a list of named expressions — the metrics, filters, and derviations that consumers should reuse rather than reinvent. Where a [constraint](#table-constraints) states something that must hold of the data, a definition states something that can be computed from it.
```
simonpcouch:
```suggestion
A table's `definitions` property is a list of named expressions — the metrics, filters, and derivations that consumers should reuse rather than reinvent. Where a [constraint](#table-constraints) states something that must hold of the data, a definition states something that can be computed from it.
```

site/spec.md
```diff
@@ -353,6 +354,36 @@ constraints:
...
+
+Each entry is a map with:
+
+* `name` (required): the definition's name. Must be non-empty and unique within the table.
+* `expr` (required): an expression in the [expression language](expressions.md). Unlike an assertion, it need not be boolean.
+* `description`: human-readable description of what the expression computes and when to use it.
```
simonpcouch: Noting that there's not a `type` here, so the only way to derive the type of an entry is to use the CLI (or duplicate the logic to parse that expression language). My initial reaction is that asks too much of downstream consumers like posit-dev/commons—in order to figure what sort of thing this is, we'd need users to have the data-dict CLI installed so that we could parse the expressions.

That said, commons' current use case would possibly need the CLI too, now, since we'd need to translate the `expr`s for the agents to use them directly in SQL. I think this makes the 'it's just YAML' interoperability story a bit more difficult.

simonpcouch: Do you have plans to make the data-dict CLI available as an R package? If that's the case, commons could ship a gnarly port of the expression translation code in the short term and then take on a dependency on that package later on.

simonpcouch (commented):
End-of-day brain is pretty fried, but some initial reactions🍳

site/spec.md
hadley: Yeah, requiring the user supply a `type` (or data-dict add one automatically) doesn't buy you a lot because you also won't get any of the transformations. 

In the short-term, you could also assume that the end-user has called `data-dict export` to create a self-containing json file which you could then use. I think the JSON will be a bit less appealing for an agent to read, but it's obviously much easier to compute with, and includes everything that `data-dict` knows about the data.

hadley: The export is documented at https://data-dict.tidyverse.org/export.html and you can (now!) see examples of generated JSON at https://data-dict.tidyverse.org/examples/.

hadley: And you might want to look at #204 too, since that shows what assertion translations look like.

simonpcouch: > In the short-term, you could also assume that the end-user has called data-dict export to create a self-containing json file which you could then use.

I'm hesitant to make the argument that a user ought to install a CLI (requiring Cargo—there are not yet prebuilt binaries, to my knowledge) in order to use the R package.

gaborcsardi (approved):
I think we'll need to say something about validation as well. But we can do that later of course.

Do we need to say something about when `COLUMNS()` is allowed in a definition?

We say "a single value for a group of rows", but we don't have a `GROUP BY`, so this is a bit weird. 

Should we say more about name collisions with column names?

Maybe allow `label` and `details`, so we have the same fields everywhere?

site/spec.md
hadley: That's fair but will resolve in the near future.

simonpcouch: Just to be sure I understand your direction here—which piece will resolve? There will be prebuilt binaries, there will be an R wrapper package, or something else?

hadley: Both

hadley:
For validation, I suggest we validate that:

* Names are unique, and unique within columns
* Definitions can use other definitions, but no cycles allowed
* They are valid expressions (i.e. yield unambiguous types)
* Error if use of `COLUMNS()` yields anything other than a boolean (might already be true?)

For label/details, I was thinking that definitions belong to the same class as `relationships` and `constraints`, i.e. they have `expr` and `description` but not labels and details. OTOH `relationships` and `constraints` don't have names, so maybe `definitions` are closer to `columns`. I think I'm convinced.

hadley:
@gaborcsardi could you please take another quick look now that I've implemented validation?

gaborcsardi (approved):
Oh, I only meant adding it to the validation spec, for this PR. Nevertheless it looks good.

gaborcsardi:
One issue: when a definition fails to parse, all other definitions referring to it fail to parse as well. This is not great, and a failing definition should probably still register as ok and not trigger further failures.

gaborcsardi:
Claude found another issue: assertion referencing a definition is silently dropped at data/translate level. I.e. if a constraint references a definition, that constraint is always satisfied and it is never translated.

gaborcsardi:
Seems like we can get duplicate `COLUMNS()` in an expression if one (or more) of them is coming from a definition.

gaborcsardi:
Btw. maybe `definitions` should also allow `todo` like normal columns?

hadley:
All fixed! I also extracted out a tree walker which simplifies a bunch of repeated code.

PR #203, opened 2026-08-10
Define `filter (where ...)` clause

hadley:
Came up as part of #134

This is the equivalent of `sum(order_total[status_cd == 90])`


site/expressions.md
```diff
@@ -361,6 +361,17 @@ An aggregate folds a column over every row of the table into a single value, so
...
+#### Filtering: `FILTER (WHERE ...)` {#filter}
+
+Any aggregate may be narrowed to a subset of rows with a `FILTER (WHERE ...)` clause, written after the closing parenthesis:
+
+```yaml
+- name: net_revenue
+  expr: SUM(order_total) FILTER (WHERE status_cd = 90)
+```
+
+The condition is a `row`-shape boolean expression over the same table, and only rows where it is true are folded; following [CHECK semantics](#truth-and-null), a row whose condition is null is excluded, exactly as `false` is. A filter that admits no rows leaves the aggregate with [empty input](#empty-input), so `SUM(x) FILTER (WHERE false)` is null and `COUNT(x) FILTER (WHERE false)` is 0.
```
gaborcsardi: Small contradiction: `WHERE false` is actually not row-shape.

site/expressions.md
```diff
@@ -361,6 +361,17 @@ An aggregate folds a column over every row of the table into a single value, so
 
 Their signatures are written over the [type classes](#type-classes).
 
+#### Filtering: `FILTER (WHERE ...)` {#filter}
+
+Any aggregate may be narrowed to a subset of rows with a `FILTER (WHERE ...)` clause, written after the closing parenthesis:
```
gaborcsardi, line 367: I guess `FILTER` and `WHERE` are case insensitive. Also, we need to add them as reserved words?

gaborcsardi (approved):
Mostly straightforward, at least at the spec level. 

I think we don't actually need to make `FILTER` and `WHERE` a reserved word, but I guess we might as well.

The interaction with `COLUMNS()` is not explicitly defined, we should probably say something about that.

We'll need to add some check spec to validation.md for `FILTER`.

site/expressions.md
gaborcsardi: This is not `CHECK` semantics, I think. it is `WHERE` and `HAVING` semantics in SQL. `CHECK` semantics allows nulls.

hadley:
You're imagining that you could use `COLUMNS()` in `WHERE()`? Like `ROW_COUNT() FILTER (WHERE COLUMNS(*) IS NOT NULL)`? 

Although that does not do what I expect in duckdb (from claude)

```sql
CREATE TABLE t AS SELECT * FROM (VALUES (1,10,100),(2,NULL,200),(3,30,NULL)) v(x,a,b);

SELECT sum(x) FILTER (WHERE COLUMNS(*) IS NOT NULL) FROM t;
-- ┌───┬───┬───┐
-- │ x │ a │ b │
-- ├───┼───┼───┤
-- │ 6 │ 4 │ 3 │
```

gaborcsardi:
I think we can make `ROW_COUNT() FILTER (WHERE COLUMNS(*) IS NOT NULL)` do what you expect by simply expanding `COLUMNS()` within the condition and `AND`-ing them, as usual. I guess this just works. We still have the at most one `COLUMNS()` per expressions of course.

So technically nothing is needed, although an example may be still nice.

hadley:
@gaborcsardi now with implementation

site/expressions.md
```diff
@@ -362,6 +362,24 @@ An aggregate folds a column over every row of the table into a single value, so
...
+
+```yaml
+- name: net_revenue
```
gaborcsardi, line 370: This needs #202 to be a valid example. Just noting.

site/expressions.md
```diff
@@ -362,6 +362,24 @@ An aggregate folds a column over every row of the table into a single value, so
...
+  expr: SUM(order_total) FILTER (WHERE status_cd = 90)
+```
+
+You can use `COLUMNS()` in the filter clause, with its usual interpretation of `AND`ing together the results from all selected columns:
```
gaborcsardi, line 374: I think the example is wrong here, it does not do what you want if you `AND` together the results from all selected columns. We could be more explicit and say that the aggregate is folded once per selected column, and use a different example. E.g:

```
`COLUMNS()` may appear in the filter clause, and [the usual rule](#selecting-multiple-columns) applies: the whole expression — the fold included — is evaluated once per selected column, so the `AND` sits outside the aggregate, not inside the condition.

- assert: AVG(qty) FILTER (WHERE COLUMNS('flag_.*')) > 0
  description: Mean quantity is positive among the rows each flag marks.
```

crates/data-dict/src/emit/r.rs
```diff
@@ -643,34 +663,43 @@ fn write_func(d: Dialect, cx: &mut Ctx, op: Op, args: &[TypedExpr]) -> Result<()
...
-                    why: "a bare predicate has no row count in base R; \
-                          use `nrow(t)` where the predicate is applied",
-                });
+        Op::RowCount => match filter {
+            // `n()` and `.N` take no subset, so count the rows the
+            // condition keeps.
+            Some(filter) => {
+                cx.push("sum(");
+                cx.free(filter)?;
+                cx.push(", na.rm = TRUE)");
```
gaborcsardi, line 689: This is going to break on constant conditions, e.g. `ROW_COUNT() FILTER (WHERE TRUE)` is translated to `sum(TRUE, na.rm = TRUE)`, which is always 1.

I guess we'd need to add a `rep_len()` with length `n()` / `.N` here, and it would be unsupported in base R.

```
Op::RowCount => match filter {
    Some(f) if f.shape == Shape::Const => {
        // A constant condition recycles, so it needs the row count to
        // count against.
        cx.push("sum(rep_len(");
        cx.free(f)?;
        match d {
            Dialect::Tidyverse => cx.push(", n())"),
            Dialect::DataTable => cx.push(", .N)"),
            Dialect::Base => return Err(Unsupported { .. }),
        }
        cx.push(", na.rm = TRUE)");
    }
    Some(f) => { /* current code */ }
    None => { /* current code */ }
}
```

site/validation.md
```diff
@@ -50,7 +50,7 @@ A validator reports two severities of problem: **errors** and **warnings**. The
...
+| S21 | Ill-typed assertion | E | An `assert` expression is syntactically valid but semantically wrong: an operator or function applied to the wrong operand type (including a column a `COLUMNS(...)` selects, and an argument outside a signature's [type class](expressions.md#type-classes), such as `SUM` of a string), a wrong function arity, a non-boolean top-level expression, a `FILTER (WHERE ...)` clause on a non-aggregate function or with a non-boolean condition, more than one `COLUMNS(...)`, a malformed `SIMILAR TO` / `COLUMNS('...')` regex, a field access on anything but a `struct` (including through a `list`), or a bare `struct` or `list` column used where a value is needed (anywhere but `IS [NOT] NULL` and `COUNT`, which ask only whether a value is null). |
```
gaborcsardi, line 53: At least for `NOW() FILTER (...)` we do not report S21, but the `NOW()` is parsed and then S19 is reported for the malformed expression. (I am not saying that this is super important, though...)

site/validation.md
```diff
@@ -59,7 +59,7 @@ A validator reports two severities of problem: **errors** and **warnings**. The
...
 | S29 | Invalid constraint on list or struct | E | A `primary_key`, `foreign_key`, or `unique` constraint appears on a `list` or `struct` column. (There is no such check for fields: a field can't carry `constraints` at all, which the schema enforces structurally.) |
-| S30 | Nested aggregate | E | An aggregate function's argument contains another aggregate call, as in `AVG(MIN(x))`: an aggregate folds one value per row, and another aggregate gives it a single value. This is the only [shape](expressions.md#shapes) an expression can get wrong — every other combination is legal, including mixing a row-level subexpression with an aggregate one (`value <= 2 * MIN(value)`). |
+| S30 | Nested aggregate | E | An aggregate function's argument contains another aggregate call, as in `AVG(MIN(x))`: an aggregate folds one value per row, and another aggregate gives it a single value. The same goes for a `FILTER (WHERE ...)` condition. These are the only ways an expression's [shape](expressions.md#shapes) can be wrong — every other combination is legal, including mixing a row-level subexpression with an aggregate one (`value <= 2 * MIN(value)`). |
```
gaborcsardi, line 62: We need to change the message in validate_spec.rs to reflect this change. Currently it is 
```
            "S30" => "An aggregate can't be nested inside another aggregate.",
```
which does not cover the new failure mode.

PR #204, opened 2026-08-10
Export assertion translations

hadley:


site/export.md
```diff
@@ -101,10 +101,29 @@ Both `references` and `referenced_by` are derived from `relationships`, not read
...
+{
+  "target": "SQL(duckdb)" | "R(tidyverse)",
+  "code?": "string",              // the predicate, absent when the target refused
+  "error?": "string",             // why the target refused, absent when it produced code
+  "notes?": ["string", ...]       // edge cases where the translation diverges from the
+                                  // expression's own semantics; absent when it agrees exactly
+}
+```
+
+`notes` flags constructs whose translation agrees with the expression except on a documented edge (e.g. integer division overflow, division by zero) — each note names the edge, so a consumer can surface the caveat alongside the code.
```
gaborcsardi: I don't think we have integer division overflow because we don't have integer division at all. :)

gaborcsardi (approved):

PR #208, opened 2026-08-11
Expressions: implement infinity and NaN

gaborcsardi:
- Update spec to include infinity and NaN, including two new
  literals.
- New IS_FINITE(), IS_INFINITE(), IS_NAN().
- Update execution spec.
- MOD(x, 0) is NaN.
- Update ordering: -INF < finite < INF < NaN. All NaN one value.
- D10 not needed any more.
- NaN rejected as a bound.

For now this is the spec only, no code changes.


crates/data-dict/tests/snapshots/validate_spec__s12_nan_is_not_a_bound.snap
```diff
@@ -0,0 +1,25 @@
...
+$learn_more: http://data-dict.tidyverse.org/
+tables:
+  - name: table
+    columns:
+      - name: mass
+        type: number(quantity)
+        units: kg
+        range: [.nan, 10]
+────────────────────────────────────────
+error[S12]: Each `range` value must have a place on the number line.
```
hadley, line 14: Every?

hadley (approved):

crates/data-dict/tests/snapshots/validate_spec__s12_nan_is_not_a_bound.snap
gaborcsardi: 'Each' meaning 'each of the two'. Should this still be 'Every'? Btw. there are a lot of instances of this. Should we change all of them? E.g.

```
Each `examples` value of a `string` column must be a string.
Each `range` value of a `list(list(number(quantity)))` column must be a number.
Each `examples` value must have a place on the number line.
Each `examples` value of a `number(id)` column must be a number.
...
```

hadley: Oh lets do it in a separate PR then: https://github.com/tidyverse/data-dict/issues/211

PR #212, opened 2026-08-11
Export definitions, with translations

### LINKED ISSUE #134: Define metrics/filters/examples

hadley:
A table's `definitions` now appear in the JSON export (`export-spec`/`export-data`), documented in `site/export.md`:

- Each definition exports with its `kind` (`filter`/`metric`/`derived`, read off the expression's resolved type and shape), its value `type`, and its references resolved — `columns` it reads and `definitions` of the same table it builds on.
- Definitions carry `translations` like assertions do. A reference to another definition renders as a bare name, as if it were a column (the IR lowerer resolves it through the environment's definition types), so substituting the referenced definition's own translation is the client's job. Assertions referencing definitions get the same treatment, and assertion `columns` no longer misreport a referenced definition as a column (new `definitions` key). The `translate` CLI checks and lowers against the same definition-aware environment.
- New `resolve_definitions` in `validate_spec.rs` is a silent variant of the validation-time resolution, reused by the export and the CLI.

Tests: targeted assertions plus a full-document snapshot (`export__definitions_export_snapshot.snap`) showing the shape end to end. Full workspace suite, clippy, and fmt pass.

Open question flagged for follow-up: filter translations use assertion `CHECK` semantics (null passes), which may not suit row-selection use.

Fixes #134

### DIFF
--- a/crates/data-dict/src/validate_spec.rs
+++ b/crates/data-dict/src/validate_spec.rs
@@ -429,42 +429,83 @@ fn run_assertion_check(
         out,
     );
 
-    // The at-most-one `COLUMNS(...)` rule binds the expression as it
-    // evaluates, with definition references substituted in: an assertion and
-    // a filter it references may each use one, which combines to two. Report
-    // each reference that brings a selection in over the budget of one. (More
-    // than one in the assertion's own text was already reported above.)
+    check_expanded_selections(table, expr, None, &assertion.text, enclosing, out);
+}
+
+/// The `COLUMNS(...)` rules bind the expression as it evaluates, with
+/// definition references substituted in: a selection travels to whatever
+/// references the definition holding it. Report each reference that brings a
+/// selection in over the budget of one, and — for a definition (`ty` is
+/// `Some`), where a selection is meaningful only in a boolean filter — a
+/// reference that brings one into a non-boolean expression. Selections in the
+/// expression's own text were already checked by `analyze`.
+fn check_expanded_selections(
+    table: &Table,
+    expr: &assert_expr::AssertExpr,
+    ty: Option<assert_expr::Ty>,
+    text: &Spanned<String>,
+    enclosing: &[&Spanned<String>],
+    out: &mut ProblemSet,
+) {
     let own = assert_expr::columns_selection_count(expr);
-    if own <= 1 {
-        let defs = definition_exprs(table);
-        let with_selection: std::collections::HashSet<&str> = defs
-            .iter()
-            .filter(|(_, e)| assert_expr::columns_selection_count(e) > 0)
-            .map(|(name, _)| name.as_str())
-            .collect();
-        let mut refs: Vec<(usize, usize)> = Vec::new();
-        assert_expr::visit(&expr.root, |e| {
-            if let assert_expr::ExprKind::Column(path) = &e.kind
-                && path.len() == 1
-                && with_selection.contains(path[0].as_str())
-            {
-                refs.push((e.start, e.end));
-            }
-        });
-        // The first selection is the allowed one: the assertion's own if it
-        // has one, else the first reference's.
-        for &(start, end) in refs.iter().skip(1 - own) {
-            let span = subspan(&assertion.text.span, start, end)
-                .unwrap_or_else(|| assertion.text.span.clone());
+    if own > 1 {
+        return;
+    }
+    let defs = definition_exprs(table);
+    let with_selection: std::collections::HashSet<&str> = defs
+        .iter()
+        .filter(|(_, e)| assert_expr::columns_selection_count(e) > 0)
+        .map(|(name, _)| name.as_str())
+        .collect();
+    let mut refs: Vec<(String, usize, usize)> = Vec::new();
+    assert_expr::visit(&expr.root, |e| {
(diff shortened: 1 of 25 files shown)


crates/data-dict/tests/snapshots/export__definitions_export_snapshot.snap
```diff
@@ -0,0 +1,150 @@
...
+            "tile_size"
+          ],
+          "translations": [
+            {
+              "target": "SQL(duckdb)",
+              "code": "\"tile_size\" IN ('Mid-Market-3', 'Enterprise-1')"
+            },
+            {
+              "target": "R(tidyverse)",
+              "code": "is.na(tile_size) | (tile_size %in% c(\"Mid-Market-3\", \"Enterprise-1\"))"
```
hadley, line 130: This implements `check` semantics which might be wrong for definitions?

crates/data-dict/tests/snapshots/export__definitions_export_snapshot.snap
```diff
@@ -0,0 +1,158 @@
...
+          "columns": [
+            "order_total"
+          ],
+          "definitions": [
+            "is_enterprise"
+          ],
+          "translations": [
+            {
+              "target": "SQL(duckdb)",
+              "code": "sum(CASE WHEN \"is_enterprise\" THEN \"order_total\" ELSE 0 END)",
```
hadley, line 156: I'm assuming that the caller is responsible for inlining the other definitions.

crates/data-dict/tests/snapshots/export__definitions_export_snapshot.snap
gaborcsardi: I don't think it is wrong. At least surely not when the definition is used in a constraint, right?

crates/data-dict/src/validate_spec.rs
```diff
@@ -658,6 +667,66 @@ fn check_definition_exprs(table: &Table, out: &mut ProblemSet) -> HashMap<String
     resolved
 }
 
+/// Resolve every referable definition's type and shape in dependency order,
+/// discarding findings. For export, which runs only after spec validation
+/// has passed, so there is nothing to report; a definition that can't be
+/// resolved (a reference cycle, already reported as S33) maps to the
+/// permissive `Ty::Any`, as it does during validation.
+pub(crate) fn resolve_definitions(table: &Table) -> HashMap<String, DefType> {
```
gaborcsardi, line 777: IDK if it is worth doing anything about this, but I believe validation already computes much of this. So we could keep that around.

gaborcsardi (commented):
Seems like definitions are reported as columns for `translate`:

```yaml
$version: "0.1.0"
description: Each row is an order. All orders are included.
tables:
  - name: orders
    columns:
      - name: tile_size
        type: string
        examples: ["Enterprise-1"]
      - name: order_total
        type: number(quantity)
        units: usd
        range: [0, .inf]
    constraints:
      - assert: is_enterprise      # references the definition
    definitions:
      - name: is_enterprise
        expr: tile_size IN ('Enterprise-1')
```

```
❯ cargo run translate /tmp/example.yaml | jq
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/data-dict translate /tmp/example.yaml`
[
  {
    "expr": "is_enterprise",
    "table": "orders",
    "type": "boolean",
    "columns": [
      {
        "table": "orders",
        "column": "is_enterprise"
      }
    ],
...
```

hadley:
We might more to think more deeply about how to align `translate` with export? Maybe it's `translate` is just something carved out of export? Or maybe `translate` can just be some flags to `export` now? Anyway, we can iterate on that elsewhere.

PR #215, opened 2026-08-12
Translate expressions to R(base) and R(data.table)

hadley:
## Summary

The spec lists three R dialects (`base`, `tidyverse`, `data.table`) but only `R(tidyverse)` was implemented. This generalises the tidyverse emitter into one R emitter parameterised by dialect (`emit/r.rs`, replacing `r_tidyverse.rs`) and registers `R(base)` and `R(data.table)`, so `--target r` now resolves to `R(base)`, the spec's family default.

All three dialects share R's core semantics (three-valued logic, `%%`, the `%in%` null guard, date promotion, NaN notes); they differ only in spelling:

| Construct | `R(tidyverse)` | `R(base)` | `R(data.table)` |
|---|---|---|---|
| `LENGTH` / `LOWER` / `TRIM` | `str_length` / `str_to_lower` / `str_trim` | `nchar` / `tolower` / `trimws` | as base |
| `LIKE` prefix/suffix | `str_starts(x, fixed(p))` | `startsWith(x, p)` | as base |
| `SIMILAR TO` | `str_detect` (ICU note) | guarded `grepl(…, perl = TRUE)` (PCRE note) | via `fifelse` |
| `CASE` | `case_when` | nested `ifelse` | `fcase` |
| `BETWEEN` | `between()` | `x >= lo & x <= hi` | `between()` |
| `COLUMNS(...)` | `if_all(...)` | expanded conjunction | expanded conjunction |
| `ROW_COUNT()` | `n()` | **refused** (no row count in a bare base-R predicate) | `.N` |
| `COUNT_DISTINCT` | `n_distinct(x, na.rm = TRUE)` | `length(unique(x[!is.na(x)]))` | `uniqueN(x, na.rm = TRUE)` |
| kind tests | `if_else` | `ifelse` | `fifelse` |

## Notes on the choices

- **Null guards:** verified in an R session that `startsWith`/`endsWith` propagate `NA` natively, so only `grepl` (which returns `FALSE` for an `NA` subject) carries an `ifelse(is.na(x), NA, …)` guard.
- **data.table selections:** the `.SD`/`Reduce(`&`, lapply(...))` idiom is a whole query, not a self-contained predicate expression, so per the spec's criterion data.table expands to a conjunction.
- **Spec sync:** the spec already defined these targets; the only spec edit is aligning the example output with the emitters' integer literals (`10L`). No new feature wording, so no sign-off gate applies.
- **Tests:** the R emitter tests moved from `emit/mod.rs` into `emit/r.rs`, reorganised by construct with the dialect variants side by side where they differ. Export snapshot updated (now carries all four targets). Full workspace tests, clippy, and fmt are clean.


gaborcsardi (approved):
Cool! It is a pity that we cannot do `ROW_COUNT()` in base R, but I don't see what we could do about it.

PR #218, opened 2026-08-12
Enforce `display: restricted` in export, validate-data, and render

### LINKED ISSUE #206: `display: restricted` is not enforced by render/describe/validate-data
## Summary

The spec says a `display: restricted` column "must be excluded from default user interfaces and other user-facing output, including tables, plots, and downloads" (site/spec.md). In practice, `render`'s profiling embeds every distinct value of a restricted column in the output HTML anyway.

## Repro

`data-dict.yaml`:
```yaml
$version: 0.1.0
description: repro
tables:
  - name: t
    source:
      parquet: t.parquet
    columns:
      - name: secret
        type: string
        display: restricted
        examples: ["PLACEHOLDER"]
        description: should never surface real values
```

`t.parquet`: a `secret` column with real/sensitive-looking string values.

```
data-dict render data-dict.yaml -o out.html
grep -o '"secret".\{0,400\}' out.html
```

**Expected:** the `secret` column's real values are excluded from the profile (or the whole `profile` key is omitted for that column).

**Actual:** `out.html` contains the column's full value list verbatim, in both `sample_values` and `common_values` (every distinct value, with its count).

## Where this lives in the source (checked at `ffd3de3c`)

- `display` is parsed and schema-validated (`crates/data-dict/src/model.rs:149-150`, only `"restricted"` accepted) but never read again after that.
- `crates/data-dict-parquet/src/profile.rs` and `.../describe.rs` — the modules that build `sample_values`/`common_values`/`histogram` — have no `display`/`restrict` references at all.
- `crates/data-dict/src/export.rs:610-660`, `build_column()`: `display: col.display.clone()` is written straight into the output JSON as inert metadata, alongside an unconditionally-attached `profile`. The only profile-gating present is unrelated (dropping samples for a column that already declares an exhaustive `enum` `values` set).
- `run_describe()` (`crates/data-dict-cli/src/main.rs:236`) takes a bare `.parquet` path with no dictionary in scope, so `data-dict describe` can't know a column is restricted even in principle.
- The frontend (`crates/data-dict-cli/render/*.js`) has zero "restricted"-related code — no badge, no masking.
- The only place this is actually realized is `crates/data-dict-cli/skills/read-data-dict.md:50-55`, which asks *AI agents* to self-censor restricted columns when reading the dictionary — this doesn't help a human opening the rendered HTML, or any tool not using that specific skill prompt.

## Suggested fix

Gate profile construction (or drop `sample_values`/`common_values`/`histogram`, keeping e.g. `missing`/`distinct.count`) in `export.rs`'s `build_column()` whenever `col.display == Some("restricted")`. `describe` is a harder case since it doesn't take a dictionary today — possibly out of scope for this specific fix, or grounds for an optional `--dict` flag. Might also be worth a small frontend affordance (badge/masking) so a `display: restricted` column reads clearly as such even if its raw values were ever present.

## Why this matters

The spec's own wording — "must be excluded... including tables, plots, and downloads" — reads as a requirement of the tool itself, not just guidance for downstream consumers. Right now the only safe way to use `render`/`describe`/`validate-data` on a dictionary with PII columns is to point `source` at a copy of the data with those columns already redacted upstream, which somewhat defeats the purpose of having `display: restricted` as a dictionary-level safeguard at all.

hadley:
Fixes #206.

A `display: restricted` column's real data values no longer appear in tool output:

- **export**: `ExportProfile::restrict()` strips `sample_values`/`common_values`/`range`/`histogram` from a restricted column's profile, keeping only counts (`missing`, `distinct`). Website rendering goes through this export, so restricted values can no longer reach the rendered HTML (verified with the issue's repro: zero occurrences of the source values in the output).
- **validate-data**: D04/D05 diagnostics withhold offending-value samples — both the rendered message ("values restricted") and the structured `ProblemKind::values` — for restricted columns.
- **render UI**: restricted columns get a `restricted` badge next to the column name, styled like the PK/FK chips.

Also:
- `site/export.md` documents that restricted profiles are counts-only, so consumers can rely on the export never containing data values.
- `AGENTS.md` gains a "Restricted columns" section stating the invariant and pointing future work at the two enforcement points.
- The `create-data-dict` skill now flags likely-PII columns, marks them `display: restricted`, and replaces real draft-generated `examples` with plausible fakes.

Out of scope: `describe` takes a bare parquet path with no dictionary, so it can't know a column is restricted — a possible future `--dict` flag.


hadley:
@Laszlo75 would you mind taking a quick look at this? The goal is to both stop the existing leakage of restricted values, and to hopefully put some guardrails in place to make sure we don't accidentally do it again in the future.

Laszlo75:
Hi @Hadley !
Tested it, works well. Restricted variables are not sampled in the rendered output anymore. Thanks for looking into this!

PR #226, opened 2026-08-13
Document JSON validation report format

### LINKED ISSUE #222: Documented JSON validation output (spec)
Cf. https://github.com/tidyverse/data-dict/issues/126#issuecomment-5269474159

We don't need to include the same problem information in multiple ways, the HTML generator / VS code extension / etc. can reshape a JSON file very effectively. So we just need to have a list of problems, like we already do have now, possibly with some extended information, and we also need to document the output format.

gaborcsardi:
Needs:
- [x] #227

Closes #222.


gaborcsardi:
Still updating this to make sure that it has everything needed for the HTML report.

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+
+Which steps exist follows from the dictionary: a `required` column gets a `D01` step, a `unique` one a `D02` step, each `assert` a `D07` step, and a column with no constraint gets none. A spec check (`S##`) reads the document as a whole rather than one declared target, so it is reported through `problems` alone, and `steps` is empty for a spec-level run.
+
+Every step the level attempted is listed, including the ones it could not weigh. A step is missing only when its level never ran at all: a spec error stops the run, so the metadata and data steps that would have followed are absent rather than listed unevaluated.
+
+### Problem
+
+```jsonc
+{
+  "code?": "D04",                // the check's code in validation.md
```
hadley:
```suggestion
  "code": "D04",                // the check's code in validation.md
```

?

site/report.md
```diff
@@ -0,0 +1,299 @@
+# Report
+
+A **report** is one validation run's findings as JSON, so a program can act on them. Every check that has a code in [validation.md](validation.md) (`S##`, `M##`, or `D##`) reports through this one document, as does a failure that stopped the run before any check could start. The `data-dict` CLI writes one for any validation run.
+
+A report is a superset of the diagnostics rendered for a person: every position a diagnostic highlights is in it, and it names more offending rows.
+
+The [level](validation.md#three-levels-of-validation) validated determines which checks ran, not the shape of what they report: a data-level report carries `S##` and `M##` problems too, since each level implies the ones before it. If an earlier level finds an error, the run stops there and reports only what it got to. A run that finds nothing still produces a report, with an empty `problems` list.
```
hadley: But there's something different about the "D" problems since they have associated rows, not just columns. Maybe that's worth pointing out explicitly?

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+
+### Top level
+
+```jsonc
+{
+  "$version": "0.1.0",           // version of the report document format itself
+  "status": "ok" | "warning" | "error",
+  "steps": [ Step ],
+  "problems": [ Problem ]
+}
```
hadley, line 25: Why aren't these paired? Like `[(Step, Problem)]`?

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+
+### Step
+
+```jsonc
+{
+  "id": 3,                       // 1-based, unique within this report
+  "code": "D04",                 // the check's code in validation.md
+  "table": "otters",             // the table the step checked
+  "column?": "status",           // the column, dotted for a struct field
+  "columns?": ["site", "day"],   // the key's columns, for a composite-key step
```
hadley: Unify? 

I assume for an assertion like `start_date < end_date` you'd want to report both?

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+`suggestion` is a concrete edit: splice `replacement` into the source over the suggestion's own `location`, which is an insertion point when it is empty.
+
+`step` is absent for a problem no step accounts for: a pre-flight failure, a spec problem, and an undocumented column (`M03`). A step with a `fail` above zero always has at least one problem pointing back at it.
+
+### Step
+
+```jsonc
+{
+  "id": 3,                       // 1-based, unique within this report
+  "code": "D04",                 // the check's code in validation.md
```
hadley, line 70: BTW I've been wondering if we should switch to IDs like `data-unique` or `meta-range-invalid`.

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+| `D02` | `unique` column, and the primary key as a whole | `D02`, `D03` |
+| `D04` | `enum` column, including a nested one | `D04` |
+| `D05` | single-column `foreign_key` | `D05`, `D06` |
+| `D07` | `assert` expression | `D07`, `D08`, `D09` |
+
+: {tbl-colwidths="[15,40,45]"}
+
+`M03` is the one check with nothing to declare it: the column exists only in the data, so no step covers it and its problem has no `step`.
+
+`units` is how many things the step weighed: the table's rows for a row-level data check, and `1` for a check with a single verdict — an aggregate assertion, or any metadata check. `fail` is how many of them failed, and is `0` for a step that passed. Both are absent when `evaluated` is false.
```
hadley: I think it would be clearer to just report rows? So then an aggregate assertion just reports all or nothing.

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+| `schema` | — | |
+| `parquet` | — | |
+| `table_not_found` | — | `available` |
+| `spec` | `S01`–`S31` | |
+| `type_mismatch` | `M01` | `declared`, `actual` |
+| `missing_in_data` | `M02` | |
+| `extra_in_data` | `M03` | `actual` |
+| `missing_source` | `M04` | |
+| `unreadable_source` | `M05` | |
+| `nulls_in_required` | `D01` | `count`, `rows` |
```
hadley, line 137: Like wouldn't these (with a little light human editing) be more useful than the numeric IDs?

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+
+* `count` is the **exact** total number of offending rows. It is never capped.
+* `rows` is a **sample** of the offending row numbers, ascending. Row numbers are 1-based and absolute within the table's Parquet file, so they survive row-group boundaries and can be used to seek back into the data.
+* `values` (`D04`, `D05`) are the **distinct** offending values, in first-seen order. Because they are deduplicated, `values` is generally shorter than `rows` and its entries do not line up with it.
+
+`samples` (`D07`) is the exception that does line up: one entry per row in `rows`, in the same order, giving that row's relevant columns as an object keyed by column name.
+
+```jsonc
+"rows": [12, 91],
+"samples": [{"weight": "-1.5", "length": "60"}, {"weight": "0", "length": "74"}]
```
hadley: Is this supposed to be values?

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+The four kinds with no code are pre-flight failures: the file couldn't be read (`io`), isn't YAML (`parse`), doesn't match the schema (`schema`), or a Parquet file failed mid-read (`parquet`). `spec` covers every semantic spec check, whose code varies per check.
+
+`values_outside_enum` and `assertion_violated` also arise for a nested column: `column` is then the dotted path to the struct field, and its rows are the rows of the top-level column that holds it.
+
+## Counting and sampling
+
+Three keys describe how many rows broke a check and which ones:
+
+* `count` is the **exact** total number of offending rows. It is never capped.
+* `rows` is a **sample** of the offending row numbers, ascending. Row numbers are 1-based and absolute within the table's Parquet file, so they survive row-group boundaries and can be used to seek back into the data.
```
hadley: Maybe capped instead of sample? Sample implies stochastic to me, whereas we can just truncate to cap max memory usage.

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+
+A check reports what its evidence supports, and the evidence differs. These are properties of the checks themselves, not gaps to be worked around:
+
+* `D01` and `D02` can sometimes be settled from Parquet footer statistics alone, without reading a single value. That path proves how many rows offend but not which, so `count` is present and `rows` is empty.
+* `D02` has no `values`: the uniqueness pass keeps a set of keys, not their renderings. Its `rows` are the rows of the **repeat** occurrences — the row that first held a value is not the one reported as a duplicate.
+* `D07` on an aggregate assertion (`SUM(x) > 0`) reports `assertion_false`, with no `count` and no `rows`. An aggregate is a single verdict about the whole table, so no row is to blame.
+* `D09` reports a single `row`: evaluation stops at the first overflow, since everything after it is computed from a value that already went wrong.
+* `D03`, `D06`, and `D08` report that a check *didn't run*, and carry a `reason` naming the barrier rather than any row.
+
+## Restricted columns
```
hadley, line 192: 👍

site/report.md
```diff
@@ -0,0 +1,299 @@
...
+      "step": 3,
+      "severity": "error",
+      "kind": "duplicate_values",
+      "expected": "A unique column must not contain duplicate values.",
+      "message": "has 3 repeated occurrences (rows: 118, 4092, 91043)",
+      "table": "otters",
+      "column": "id",
+      "columns": ["id"],
+      "count": 3,
+      "rows": [118, 4092, 91043],
```
hadley, line 252: Why no values here?

site/report.md
gaborcsardi: This is optional because some problems do not have a code. e.g. if there is a problem with the parquet file. (Or maybe we don't need a report in that case at all?)

site/report.md
gaborcsardi: Because this is not a 1:1 mapping. We need to include steps without any problems as well, we also have problems with no steps (e.g. unable to read the parquet file).

site/report.md
gaborcsardi: I don't have strong feelings either way, both have their advantages.

site/report.md
gaborcsardi: yeah, we can do that. the validation code has a lot of 'column' but we can refactor.

site/report.md
gaborcsardi: yeah, although we (people) are going to get used to the codes as well if they use data-dict a lot, and they are shorter and easier to refer to. but I don't mind either way.

site/report.md
gaborcsardi: yeah, I agree. we'll need to add another entry then, because a failing assertion might implicate no rows at all, so we cannot mark a pass with "implicated rows == 0" any more. but that's fine, it is cleaner that way, anyway.

site/report.md
gaborcsardi: 'values' means offending single values, 'samples' are offending rows. do you want to call it 'rows'?

site/report.md
gaborcsardi: well, the uniqueness checker does not have the actual values

gaborcsardi: which we can change of course, i.e. go back to the data after the failing check and get the values

site/report.md
hadley: Yeah, in that case I'd say we can just error.

site/report.md
hadley: All though that might imply the entire row? I think I'd prefer `values` and report them rowwise

site/report.md
hadley: Yeah, I think that's worth it.

site/report.md
gaborcsardi: that'd make sense, except that by the time we fail on the parquet file we might have produced some useful stuff on the schema.

gaborcsardi: Also, we should move the quarto-yaml-validation checks to S## checks as well, they are part of schema validation, after all.

gaborcsardi:
OK, all is done, except renaming the failure codes to more meaningful names.

site/report.md
```diff
@@ -0,0 +1,313 @@
...
+
+## Counting and capping
+
+Three keys describe how many rows broke a check and which ones:
+
+* `count` is the **exact** total number of offending rows. It is never capped.
+* `rows` are the offending row numbers, ascending, **capped** at the first so many. Row numbers are 1-based and absolute within the table's Parquet file, so they survive row-group boundaries and can be used to seek back into the data.
+* `values` are the offending values themselves, in one of two shapes, decided by the `kind`.
+
+For `D04` and `D05` a value is a bare string, and the list is the **distinct** offending values in first-seen order. Because they are deduplicated, `values` is generally shorter than `rows` and its entries do not line up with it.
```
hadley: Can we use the same form here?

hadley (approved):

site/report.md
gaborcsardi: Done. `values` is now one shape everywhere: an object per offending row, keyed by column name, lined up with `rows` and not deduplicated. So `D04`/`D05` report `[{"status": "retired"}, {"status": "unknown"}, {"status": "retired"}]` instead of a distinct list of bare strings.

Issue #223, opened 2026-08-13
CLI validation report, from the JSON output


gaborcsardi:
Cf. https://github.com/tidyverse/data-dict/issues/126#issuecomment-5269474159

Think about what logging should look like: we should assume that this will be called in CI/CD so we want the logs to be useful. This might be a simple transformation of the JSON above, or possible just JSONL if structured logs are desired. (Hadley)

OTOH I am not sure if the validation failures should be communicated via logging. I think we can have a cli report that also works from the JSON output. (Gabor)

hadley:
I note logging because I talked to a team using soda, and they only looked at the validation issues via a log. So I suspect it's important.

If your job is being run in (e.g.) airflow, I'm not sure how easy it is for us to surface a file (JSON or HTML) to the user.

gaborcsardi:
> If your job is being run in (e.g.) airflow, I'm not sure how easy it is for us to surface a file (JSON or HTML) to the user.

I am not suggesting that. I suggest still creating the file and then running the CLI report on the file as part of the job, printing to stdout. Or doing this with a single `data-dict` call. I assume they would access to the stdout of the job, no? Or they only have access to (otel-type) logging?

hadley:
Oh yeah, for logging I just meant writing to stdout/stderr.

PR #233, opened 2026-08-14
Report offending values for uniqueness checks

### LINKED ISSUE #227: Keep values in uniqueness check
We either need to keep the values, or go back to the data after some problems were found.

gaborcsardi:
Mostly straightforward. Values are only stored for duplicates, so no cost if no violating rows.

Closes #227.

### DIFF
--- a/crates/data-dict/tests/snapshots/validate_data__assertion_values_withheld_when_restricted.snap
+++ b/crates/data-dict/tests/snapshots/validate_data__assertion_values_withheld_when_restricted.snap
@@ -1,0 +1,27 @@
+---
+source: crates/data-dict/tests/validate_data.rs
+---
+$version: "0.1.0"
+$learn_more: http://data-dict.tidyverse.org/
+tables:
+  - name: t
+    source:
+      parquet: data.parquet
+    constraints:
+      - assert: a > 0
+    columns:
+      - name: a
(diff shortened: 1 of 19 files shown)


crates/data-dict/tests/snapshots/validate_data__assertion_values_withheld_when_restricted.snap
```diff
@@ -0,0 +1,27 @@
...
+      - name: b
+        type: number(quantity)
+        range: [-1000000, 1000000]
+────────────────────────────────────────
+error[D07]: An assertion must hold for every row.
+  --> dict.yaml:8:17
+   |
+LL |   - name: t
+...
+LL |       - assert: a > 0
```
hadley, line 26: Fine for this PR, but I think we should probably also switch this display to something more row based, since in most cases it's going to be the data that needs fixing, not the spec.

hadley (approved):

crates/data-dict/tests/snapshots/validate_data__assertion_values_withheld_when_restricted.snap
gaborcsardi: Yeah, that's #223 essentially, I think.

Issue #224, opened 2026-08-13
HTML validation report, from the JSON output


gaborcsardi:
Cf. https://github.com/tidyverse/data-dict/issues/126#issuecomment-5269474159

Create a human focussed (beautiful) HTML page generated by asking claude to copy pointblank style with out json file, in the same way that data-dict render currently works (e.g. using preact etc). We'll want both the validation report and the step reports in a SPA. (Need to make sure to respected restricted columns here and redact the actual values). (Hadley)

PR #237, opened 2026-08-14
Update spec for HTML report

gaborcsardi:
First step for #224.

### DIFF
--- a/site/report.md
+++ b/site/report.md
@@ -1,6 +1,6 @@
 # Report
 
-A **report** is one validation run's findings as JSON, so a program can act on them. Every check in [validation.md](validation.md) (`S##`, `M##`, or `D##`) reports through this one document, and only those checks do. The `data-dict` CLI writes one for any validation run it can get started.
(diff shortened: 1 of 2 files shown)


site/report.md
```diff
@@ -35,6 +38,31 @@ Every step the level attempted is listed, including the ones it could not weigh.
...
 The problems are in no promised order beyond this: a spec problem sits at its position in the document, and a metadata or data problem in the order the run found it. A consumer that wants another order sorts the list itself.
 
+### Run
+
+```jsonc
+{
+  "dictionary": "data-dict.yaml",           // the file the run validated
+  "level": "spec" | "meta" | "data",        // the level that ran
+  "table?": "otters",                       // the one table the run covered
+  "generated_at": "2026-08-14T09:31:02Z",   // when the run finished, UTC
```
gaborcsardi, line 48: `generated_at` will make two otherwise identical reports different. Is that a problem?

hadley: I think that's ok? It's good to know exactly when it was run.

site/report.md
```diff
@@ -35,6 +38,31 @@ Every step the level attempted is listed, including the ones it could not weigh.
...
+  "tool": { "name": "data-dict",            // what produced the report
```
hadley, line 49: What are you imagining here? Is it mainly to capture the version?

hadley (approved):

site/report.md
gaborcsardi: yeah, something like that. just de-coupling the HTML generation in general

PR #239, opened 2026-08-14
HTML report

### LINKED ISSUE #224: HTML validation report, from the JSON output

gaborcsardi:
This is not fine-tuned at all, because I thought it was more important that we could start playing with it.

Also, I don't think the CLI UI is ideal, we'll need to think about the CLI subcommands as well. We probably do not want three `validate-*` subcommands, each with their own partial report, etc. 

Closes #224.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -616,6 +616,7 @@ dependencies = [
  "indoc",
  "insta",
  "parquet",
+ "serde",
  "serde_json",
 ]
 
--- a/crates/data-dict-cli/Cargo.toml
+++ b/crates/data-dict-cli/Cargo.toml
@@ -24,6 +24,7 @@ data-dict.workspace = true
 data-dict-parquet.workspace = true
 data-dict-lsp = { workspace = true, optional = true }
 clap.workspace = true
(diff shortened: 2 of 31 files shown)


hadley:
Initial thoughts:

* We can probably just merge this pretty soon, since I'll want to do a bunch of iteration on the design and Hadley <-> claude is way more efficient than Hadley <-> Gabor <-> Claude 😄 
* I wonder if this needs a different cli component name, because to me it feels like we don't really want to tie this back to the yaml, and I think what we should be logging is something one line line per failing row per check
* I think the main page could probably make checks the primary UI, with a default sorting so that failed come to the top. Then delete problems (so only accessible via the checks table), and replace it a table that just all columns for the tables with failed rows.

PR #249, opened 2026-09-08
Rename the default expression language to sql

hadley:
Hard rename of the default expression language from `data-dict` to `sql`:

- `language: data-dict` is no longer accepted; the schema enum is `[sql, r, python]` (S62 rejects the old value, no alias).
- The translation target that prints the language's own spelling becomes a dialect of the SQL family: `--target SQL(data-dict)`. Bare `sql` still means `SQL(ANSI)`.
- Spec docs updated: `validate.md`, `spec.md`, `dev-validation.md`, `expression-execution.md`.

Also fixes a pre-existing build break: `report.rs` still included `site/validation.md` after that file was renamed to `dev-validation.md`.

## Test plan

- `cargo test --workspace` passes; three insta snapshots updated (target name, S62 enum message).
- `cargo fmt` / `cargo clippy --workspace --all-targets` clean (only pre-existing `result_large_err` warnings).
- All `site/examples/*.yaml` and the playground dictionary still validate.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,7 +22,7 @@ The repo contains:
 
 ## Spec and implementation must stay in sync
 
-The spec (`site/spec.md` + validation details in `site/validation.md`) and the implementation (the crates + `schema.yaml`) are two views of the same thing and must never drift apart.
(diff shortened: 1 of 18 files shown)


hadley:
@gaborcsardi FYI

